PR #2, opened 2020-09-04
Moved source code into `fusion_engine/` subdirectory and changed YPR to degrees.

adamshapiro0:
# Changes
- Moved all files into `point_one/fusion_engine/` directory and `point_one::fusion_engine` namespace
- Renamed `Get*()` string helper functions to `to_string()` for STL consistency
- Changed `PoseMessage` YPR definition to degrees for consistency with LLA values

# Improvements
- Added `to_string()` function for `SolutionType`
- Added `std::ostream` operators for convenience


anathan (approved):
I guess it's OK we're being consistent with YPR and LLA both being in degrees. I was fine either way honestly.

PR #3, opened 2020-09-06
Changed CRC polynomial and added Python implementation.

adamshapiro0:


anathan (approved):
Python impl is good. It's worth noting it's a different generator...not a correction 🥇

adamshapiro0:
Noted and changed.

PR #4, opened 2020-09-14
Changed to 4-byte alignment, use 32-bit float for some values, and resolved Windows build and other issues.

adamshapiro0:
# New Features
- Added an 8-bit message sequence number
- Added support for Windows compilation using Microsoft Visual Studio

# Changes
- Use 4-byte alignment and packing for all message `struct`s, and use 32-bit `float` for fields that don't require precision
- Resolved compilation errors/warnings in Windows using MSVC
- Removed CRC function overloads taking `MessageHeader`
- Fixed CRC values in example data file


README.md
```diff
@@ -18,6 +18,16 @@ time, as well as processing recorded output data. Both C++ and Python are suppor
     * [Running Examples](#running-examples)
 * [Usage](#usage)
 
+### Requirements
+
+#### C++ Support
+- C++11 or later
+- CMake 3.x or Bazel 3.x
+- GCC, Clang, or MSVC
```
anathan, line 26: We should mark MSVC as preliminary until it is run by CI

adamshapiro0: Agreed. Done.

examples/generate_data/generate_data.cc
```diff
@@ -24,137 +25,148 @@ Generate a binary file containing a fixed set of messages.
...
-  header.message_type = MessageType::POSE;
-  header.payload_size_bytes = sizeof(PoseMessage);
+  MessageHeader* header = reinterpret_cast<MessageHeader*>(buffer);
+  *header = MessageHeader();
+  header->sequence_number = 0;
+  header->message_type = MessageType::POSE;
+  header->payload_size_bytes = sizeof(PoseMessage);
 
-  stream.write(reinterpret_cast<char*>(&header), sizeof(header));
+  PoseMessage* pose_message = reinterpret_cast<PoseMessage*>(header + 1);
```
anathan, line 43: while clever, I think this is a bit unclear. I'd rather us just do the more straightforward buffer + sizeof(header)

examples/message_decode/message_decode.cc
```diff
@@ -13,8 +13,12 @@ using namespace point_one::fusion_engine::messages;
 
 /******************************************************************************/
 bool DecodeMessage(std::ifstream& stream, size_t available_bytes) {
+  // Enforce a 4-byte aligned address.
   uint8_t storage[4096];
```
anathan, line 17: can we just use alignas(4) on this element and avoid this?

examples/message_decode/message_decode.cc
```diff
@@ -48,14 +52,14 @@ bool DecodeMessage(std::ifstream& stream, size_t available_bytes) {
...
         "CRC failure. [type=%s (%u), size=%zu bytes (payload size=%u bytes], "
-        "crc=0x%08x]\n",
+        "expected_crc=0x%08x, calculated_crc=0x%08x]\n",
         to_string(header.message_type).c_str(),
         static_cast<unsigned>(header.message_type),
         sizeof(MessageHeader) + header.payload_size_bytes,
-        header.payload_size_bytes, CalculateCRC(header));
+        header.payload_size_bytes, header.crc, CalculateCRC(buffer_start));
     return false;
   }
```
anathan, line 64: we should also verify the sequence number in this example

anathan (approved):
Some minor suggetions

examples/message_decode/message_decode.cc
adamshapiro0: That's a good idea, makes things cleaner.

examples/generate_data/generate_data.cc
adamshapiro0: I was previously doing something like `buffer += sizeof(MessageHeader)` but it got a bit messy with the buffer alignment since I needed to keep track of 3 pointers: storage, start of aligned storage (for CRC calc and file write), current buffer offset. `alignas()` makes it cleaner so I can go back to the more clear way.

PR #5, opened 2020-09-15
Added attitude and velocity std dev to `PoseMessage`.

adamshapiro0:
# Changes (not backwards compatible)
- Added attitude and velocity std dev
- Resized sequence number to 32b
- Renamed position std dev field and changed definition to ENU instead of ECEF
- Changed velocity definition to body frame instead of ENU

# Fixes
- Corrected missing velocity field in Python `PoseMessage` class
- Corrected `float` fields in Python messages


anathan (approved):
looks good!

src/point_one/fusion_engine/messages/solution.h
```diff
@@ -42,28 +42,40 @@ struct PoseMessage {
   double lla_deg[3] = {NAN, NAN, NAN};
 
   /**
-   * The platform attitude (in degrees), if known, described as Euler-321 angles
-   * (yaw, pitch, roll), or `NAN` if attitude has not been initialized.
+   * The position standard deviation (in meters), resolved in the local level
```
anathan, line 45: I think most people refer to this as in the Local Tangent Plane

adamshapiro0: Yeah I've seen both but that seems more popular. Will change.

anathan (approved):
looks good

PR #6, opened 2020-09-16
Added GPS time std dev and separated GNSS info/satellite messages (not backwards compatible).

adamshapiro0:
# Changes
- Separated `GNSSInfoMessage` (fixed length) and `GNSSSatelliteMessage` (variable length per satellite)
- Added GPS time std dev to `GNSSInfoMessage`
- Defined satellite `usage` field as a bit mask (currently indicating "used" or "not used")

# Fixes
- Handle unrecognized message types gracefully in Python


python/fusion_engine_client/messages/defs.py
```diff
@@ -43,6 +43,7 @@ class MessageType(IntEnum):
   # INS solution messages.
   POSE = 10000
   GNSS_INFO = 10001
+  GNSS_SATELLITE = 10002
```
anathan, line 46: would you be opposed to making this plural (since its a repeated message)

anathan: Actually I take it back, this is more consistent with the class name and it'd be kind of weird to have it be Satellites in the class.

anathan (approved):
looks good

PR #8, opened 2020-09-23
Added auxiliary pose data and IMU measurement support.

adamshapiro0:
# New Features
- Defined `PoseAuxMessage`
- Defined `IMUMeasurement`

# Fixes
- Fixed size returned by `PoseMessage::unpack()` in Python

### DIFF
--- a/python/examples/message_decode.py
+++ b/python/examples/message_decode.py
@@ -47,7 +47,13 @@ def decode_message(header, data, offset):
         print('GNSS info message @ P1 time %s [sequence=%d, size=%d B]' %
               (str(contents.p1_time), header.sequence_number, len(data)))
         print('  GPS time: %s' % str(contents.gps_time.as_gps()))
-        print('  GDOP: %.1f' % contents.gdop)
+        print('  Reference station: %s' %
+              (str(contents.reference_station_id)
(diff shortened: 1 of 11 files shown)


python/examples/message_decode.py
```diff
@@ -47,7 +47,13 @@ def decode_message(header, data, offset):
...
               (str(contents.p1_time), header.sequence_number, len(data)))
         print('  GPS time: %s' % str(contents.gps_time.as_gps()))
-        print('  GDOP: %.1f' % contents.gdop)
+        print('  Reference station: %s' %
+              (str(contents.reference_station_id)
+               if contents.reference_station_id != GNSSInfoMessage.INVALID_REFERENCE_STATION
+               else 'none'))
+        print('  Last differential time: %s' % str(contents.last_differential_time))
+        print('  GDOP: %.1f  PDOP: %.1f' % (contents.gdop, contents.pdop))
+        print('  VDOP: %.1f  HDOP: %.1f' % (contents.hdop, contents.vdop))
```
anathan, line 56: v and h are swapped here I think?

anathan (changes requested):
just check that one print, otherwise g2g

PR #10, opened 2020-10-30
Fixed missing message types in to_string() function.

adamshapiro0:
# Changes
- Print solution type in Python decoder example

# Fixes
- Added missing message types to `to_string()`


lukes3315 (approved):
Looks good

PR #11, opened 2020-12-14
Added Python tools for reading and analyzing logged data.

adamshapiro0:
# New Features
- Added `FileReader` class for reading messages from disk, with caching and indexing support
- Added `Analyzer` class for displaying data (currently supports pose data only)

# Changes
- Moved message string printouts to `str()`/`repr()` operators in message classes
- Added `MessageType.get_type_string()` helper function
- Reserved message types above 20000

### DIFF
--- a/.github/workflows/release_build.yml
+++ b/.github/workflows/release_build.yml
@@ -78,24 +78,24 @@ jobs:
 
     # Compile.
     - name: Install CMake 3.16.x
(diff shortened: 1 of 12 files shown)


ryancdougherty (approved):
Wow. This is some serious Python code. I don't see anything obvious that needs to be addressed.

PR #16, opened 2021-02-18
Added position extraction example, and cleaned up README and example applications.

adamshapiro0:


python/examples/extract_position_data.py
```diff
@@ -0,0 +1,72 @@
...
+        f.write('GPS Time (sec), Solution Type, Lat (deg), Lon (deg), Ellipsoid Alt (m)\n')
+        for message in pose_data.messages:
+            if message.solution_type != SolutionType.Invalid and message.gps_time:
+                f.write('%.3f, %d, %.8f, %.8f, %.3f\n' %
+                        (float(message.gps_time), message.solution_type.value, *message.lla_deg))
+
+    # Generate a KML file.
+    with open(os.path.join(output_dir, 'position.kml'), 'w') as f:
+        f.write(KML_TEMPLATE %
+                {'coordinates': '\n'.join(['%.8f,%.8f' % (message.lla_deg[1], message.lla_deg[0])
```
anathan, line 70: It would be nice to put 3d coordinates instead of 2d.

anathan (approved):

python/examples/extract_position_data.py
adamshapiro0: Yeah, I just wasn't sure exactly what to do. As far as I'm aware, KML doesn't support ellipsoid height, only MSL or above the ground. We only have ellipsoid in the data - no geoid undulation in the message currently. I could put the ellipsoid height so there's _some_ height, but that seems like it could be misleading.

anathan: Ya we'd have to pull in some geoid magic in python which guess is a pain.

PR #17, opened 2021-02-20
Changed p1bin file location and added a map display.

adamshapiro0:


anathan (approved):
Look good, I wish there was a nicer way to do the mapbox token bit, but I cannot think of antyhing.

adamshapiro0:
I guess it couldn't hurt to check for either `MAPBOX_ACCESS_TOKEN` or `MaboxAccessToken` env variables too (the camel case version is an older variable name, but both are supported by mapbox so we might as well support both too).

anathan:
Good call.

PR #23, opened 2021-06-01
Fixed missing include in crc.cc and changed RTK-fixed position color on the map.

adamshapiro0:


python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -230,7 +230,7 @@ def _plot_data(name, idx, marker_style=None):
                 # If there's no data, draw a dummy trace so it shows up in the legend anyway.
                 map_data.append(go.Scattermapbox(lat=[np.nan], lon=[np.nan], name=name, **style))
 
-        _plot_data('RTK Fixed', solution_type == SolutionType.RTKFixed, {'color': 'red'})
+        _plot_data('RTK Fixed', solution_type == SolutionType.RTKFixed, {'color': 'orange'})
```
anathan, line 233: if we're changing the color, should we consider making the size consistent? (5)

anathan (approved):
Minor aesthetic question, good to merge whichever way you decide.

python/fusion_engine_client/analysis/analyzer.py
adamshapiro0: Sure, we can do that

adamshapiro0: Went with 8 for both instead of 5. The 5 dots are a little small on the max zoom level supported by plotly.

PR #28, opened 2021-08-15
Removed unnecessary heap allocations in to_string() functions, and fixed offset in Python PoseMessage.pack().

adamshapiro0:


anathan (approved):
good catch on that pack

PR #34, opened 2021-09-08
Improved example data extraction scripts, and added Python library README.

adamshapiro0:


python/README.md
```diff
@@ -0,0 +1,115 @@
...
+   ```bash
+   pip install -r requirements.txt
+   ```
+3. Run any of the applications in `python/bin/` or the example code in `python/examples/`. For example, to plot results
+   from a `*.p1log` file or recorded in an Atlas log:
+   ```bash
+   python3 bin/p1_display.py /path/to/log/directory
+   ```
+
+#### Virtual Environment
```
anathan, line 47: While good advice I think this shouldn't be here it or be in an appendix. It's creates noise if you're just looking for information

python/examples/extract_position_data.py
```diff
@@ -68,26 +70,61 @@
...
+            for i in range(len(pose_data.p1_time)):
+                f.write('%.6f, %.6f, %d, %.8f, %.8f, %.3f, %d\n' %
+                        (pose_data.p1_time[i], pose_data.gps_time[i],
+                         pose_data.solution_type[i], *pose_data.lla_deg[:, i], num_svs[i]))
 
     # Generate a KML file.
-    with open(os.path.join(output_dir, 'position.kml'), 'w') as f:
+    path = os.path.join(output_dir, 'position.kml')
+    logger.info("Generating '%s'." % path)
+    with open(path, 'w') as f:
```
anathan, line 124: We should add timestamps to our KML

python/examples/extract_satellite_info.py
```diff
@@ -44,21 +44,26 @@
...
         logger.error(str(e))
-        os.exit(1)
+        sys.exit(1)
 
-    # Read pose data from the file.
+    # Read satellite data from the file.
     reader = FileReader(input_path)
     result = reader.read(message_types=[GNSSSatelliteMessage], show_progress=True)
+    satellite_data = result[GNSSSatelliteMessage.MESSAGE_TYPE]
+    if len(satellite_data.messages) == 0:
```
anathan, line 53: Is it necessary to bail out here? It looked earlier we handle the case where there is no GNSS data

anathan (approved):
Just minor feedback

python/examples/extract_satellite_info.py
adamshapiro0: I think you're thinking of a different script?

python/examples/extract_position_data.py
adamshapiro0: I'm not super familiar with how to do that. We're using a list of coordinates inside of a line string:
```
<LineString>
  <coordinates>
    -127.456,37.123
    -127.45678,37.12345
  </coordinates>
</LineString>
```
I don't think a `<coordinates>` element can be assigned timestamps for individual positions. I think you can set a `gx:TimeSpan` property on the `<LineString>` but that would just be the total range.

python/README.md
adamshapiro0: Sure, I'll move it to the bottom of the page

python/examples/extract_position_data.py
anathan: Here's a gist showing an example, but I think we can do this in a separate PR: https://gist.github.com/jsmorph/26c63d2bf84b7e2e041f

python/examples/extract_satellite_info.py
anathan: You're right!

PR #38, opened 2021-11-04
Added data time-alignment support, and resolved minor issues.

adamshapiro0:
# New Features
- Added built-in support for time-aligning messages by P1 time

# Fixes
- Fixed `pack_into()` calls when serializing messages
- Fixed relative paths in the generated `index.html` file when the output directory is not absolute
- Added execute permissions to example scripts


axlan (approved):
A few high level comments:

1. This may be a use case we don't care about, but I could see this feature causing issues if we have p1 timestamped messages that aren't generated simultaneously with the rest of the group. I guess we don't really do that for now since most messages that aren't generated from the filter state use system time.

If we did have messages like this, you'd need to separately call the file reader for the time aligned messages, and then ones you didn't want to time align? It might be worth considering making an explicit argument to allow a subset of the messages be time aligned.

2. This is the comment I mentioned before, but my personal preference would be to make these analysis stages more composable. I'm mostly looking at the `FileReader.read()`. While I think it's fine to have a centralized function that takes a ton of arguments that can do all these different steps together, I think breaking the steps up into dedicated functions that can more easily be reused has advantages. Basically, in this case it would be to encapsulate the just the file reading portion into its own function. This is mostly to make separating the dependencies a bit easier.

3. The way the time alignment works relies on having multiple copies of the full data set in memory. I've found Python will often have even more copies then I'd expect through some of these transform functions. This is not a problem per se, but at some point we may need to consider making a more streaming focused variant to process 24h logs.

adamshapiro0:
1. Yes, that is possible. For example, IMU data is asynchronous and will not be aligned to the pose data. An optional list of message types to be aligned is a good idea: anything not in the list will be left as is, and if unspecified it'll align everything.

2. Yeah, `read()` has definitely gotten quite big. I'm not sure how best to break it up. Are you just suggesting moving all of the read stuff (most of the code) into another function so that `read()` itself is short? Basically, leaving `read()` as like "read the data, time align, numpy conversion"?

3. Yeah definitely agree. This code is not really well suited to very long data sets at the moment. I think that really needs a bigger redesign to use a generator pattern or something. We'd have to consider what that would mean in terms of look-ahead and such to support time alignment, and also probably some sort of user control of read chunk sizes or something. Definitely to be considered for the future.

axlan:
For how to break up the read, I guess there's a few somewhat separate things going on:
- There's the initial prep of choosing the start time, and which messages are needed.
- There's the somewhat separate process of reading for indexed/non-index
- Then there's the decoding, align, numpy steps

Ideally this could be encapsulated into separate functions.

adamshapiro0:
Yeah, that seems reasonable. We can reorganize the function a little bit in a future PR.

PR #39, opened 2021-11-13
Added mixed binary log support, and other fixes.

adamshapiro0:
# New Features
- Added support for extracting FusionEngine data from binary files with mixed content
- Added a `--ignore-index` option
- Added message versions, consistent with the C++ definitions

# Fixes
- Fixed message `pack()` functions


axlan (commented):
Not knowing exactly how we're sequencing these changes, but there were some additional bugs I found in defs.py in this commit https://github.com/PointOneNav/fusion-engine-client-internal/commit/c76d8292fe0a3ca753a48bb8e2abae00ffed49ec . I mostly fixed them as part of the process of adding the encoder functionality, but just want to keep them in mind as you're changing related pieces.

axlan (approved):
Looks good, made some high level comments that might be worth exploring here, or in the larger set of changes.

adamshapiro0:
Yes, I included the `defs.py` packing fixes in the other PR where I added the encoder class.

I think both the "clean" command and the "from scratch" command are good ideas. "From scratch" is essentially `--ignore-index` when we're only talking about a `.p1log` file, but the new feature of extracting `.p1log` from a mixed binary file adds another layer to that so we might want to either add another option or redefine that option to take some sort of argument for "how far back should I go".

"Clean" would be nice so you can delete generated HTML files and stuff that takes up disk space. It's a little tricky for the mixed binary mode though. We can delete the index file easily, but once we create a `p1log` file, we don't necessarily know the name/extension of the original mixed binary file that got loaded.

PR #46, opened 2021-12-20
Added `FusionEngineFramer` helper class and a C++ TCP client example application.

adamshapiro0:
# New Features
- Added `FusionEngineFramer` class to help with framing and decoding incoming stream data (TCP, serial, etc.)
- Added an example TCP client C++ application (Linux only)


examples/tcp_client/tcp_client.cc
```diff
@@ -0,0 +1,157 @@
+/**************************************************************************/ /**
```
anathan, line 1: the name of the file should reflect this is Linux only (or we should consider using an OS independent library like boost sockets)

examples/tcp_client/tcp_client.cc
```diff
@@ -0,0 +1,157 @@
...
+
+/******************************************************************************/
+void HandleMessage(const MessageHeader& header, const void* payload) {
+  if (header.message_type == MessageType::POSE) {
+    auto& contents = *static_cast<const PoseMessage*>(payload);
+
+    double p1_time_sec =
+        contents.p1_time.seconds + (contents.p1_time.fraction_ns * 1e-9);
+
+    static constexpr double SEC_PER_WEEK = 7 * 24 * 3600.0;
```
anathan, line 42: this code is in 2 places. consider a static helper method or putting in defs or similar?

examples/tcp_client/tcp_client.cc
```diff
@@ -0,0 +1,157 @@
+/**************************************************************************/ /**
+ * @brief Linux TCP client example.
+ * @file
```
anathan, line 3: I think it's worthwhile expanding the description a bit here, in particular highlighting this is a "minimal" implementation and is not robust to network outages and reconnects (which we should highlight are critical for users to implement in any "real" system)

examples/tcp_client/tcp_client.cc
adamshapiro0: Ok, I'll rename. I specifically wanted to avoid bring in Boost since it is a bear of a dependency, and Boost ASIO is far from easy to understand.

examples/tcp_client/tcp_client.cc
adamshapiro0: Yeah, I'll move all of the message print stuff between this and the file decode example to a common function, similar to the Python examples.

examples/tcp_client/tcp_client.cc
adamshapiro0: Yep, good point.

src/point_one/fusion_engine/parsers/fusion_engine_framer.cc
```diff
@@ -0,0 +1,396 @@
...
+/******************************************************************************/
+ssize_t FusionEngineFramer::OnByte(bool quiet) {
+  // User-supplied buffer was too small. Can't parse messages.
+  if (buffer_ == nullptr) {
+    return 0;
+  }
+
+  // If warnings are disabled, run in quiet mode.
+  if (!warn_on_error_) {
+    quiet = true;
```
axlan, line 141: Not a big deal, but why introduce the extra variable here?

src/point_one/fusion_engine/parsers/fusion_engine_framer.h
```diff
@@ -0,0 +1,168 @@
...
+   * `buffer` must exist for the lifetime of this instance.
+   *
+   * @param buffer The framing buffer to use. Set to `nullptr` to allocate a
+   *        buffer internally.
+   * @param capacity_bytes The maximum framing buffer capacity (in bytes).
+   */
+  void SetBuffer(void* buffer, size_t capacity_bytes);
+
+  /**
+   * @brief Enable/disable warnings for CRC and "message too large" failures.
```
axlan, line 88: I think it's worth giving a brief explanation of a case where this is use (i.e., if the data stream is expected to have non-FusionEngine data, these failures are expected. )

axlan (approved):
Seems like a reasonable chunk of code to pull in to get the framer functionality included. As you mentioned it doesn't really address wider issues on how we want to share the common code, but we can refactor that in the long term.

PR #55, opened 2022-01-16
Fixed possible buffer overflow in framer for bogus headers on 32b platforms.

adamshapiro0:


src/point_one/fusion_engine/parsers/fusion_engine_framer.cc
```diff
@@ -59,6 +59,16 @@ void FusionEngineFramer::SetBuffer(void* buffer, size_t capacity_bytes) {
                << " B]";
     return;
   }
+  // Restrict the buffer capacity to 2^31 bytes. We don't expect to ever have a
+  // single message anywhere near that large, and don't expect users to ever
+  // pass in a buffer that size. Using uint32_t instead of size_t internally
+  // makes it easier to guarantee behavior between 64b and 32b architectures. We
+  // do 2^31, not 2^32, so we can use int32_t for return values internally.
+  else if (sizeof(size_t) == 8 && capacity_bytes > 0x7FFFFFFF) {
```
anathan, line 67: My strong preference is to limit this to the smaller size regardless of machine architecture.

src/point_one/fusion_engine/parsers/fusion_engine_framer.cc
```diff
@@ -189,14 +200,23 @@ p1_ssize_t FusionEngineFramer::OnByte(bool quiet) {
...
+      // header if we find the preamble randomly in an incoming byte stream.
       auto* header = reinterpret_cast<MessageHeader*>(buffer_);
       current_message_size_ =
           sizeof(MessageHeader) + header->payload_size_bytes;
       VLOG(3) << "Header complete. Waiting for payload. [message="
               << header->message_type << " (" << (unsigned)header->message_type
               << "), seq=" << header->sequence_number
               << ", payload_size=" << header->payload_size_bytes << " B]";
-      if (current_message_size_ <= capacity_bytes_) {
+      if (header->payload_size_bytes <=
```
anathan, line 218: It seems safer to check both conditions (old and new). In general I like.bounds checks to be directly on the variables indexing those bounds, not by a derivative quantity in case other logic is changed later.

src/point_one/fusion_engine/parsers/fusion_engine_framer.cc
adamshapiro0: Oh good point. I originally limited to 2^32 so this if was irrelevant on 32b, hence the size check. But then I had to limit to 31 so we could use negative returns internally and forgot to remove the size limit. Will fix.

src/point_one/fusion_engine/parsers/fusion_engine_framer.cc
```diff
@@ -189,14 +200,23 @@ p1_ssize_t FusionEngineFramer::OnByte(bool quiet) {
       // Compute the full message size. If the message is too large to fit in
       // the buffer, we cannot parse it. Otherwise, start collecting the
       // message payload.
+      //
+      // Note that while we compute the full message size here, we intentionally
+      // do the "too big" check below with the payload size. On a 32b machine,
```
anathan, line 205: I don't understand this comment, it seems to explain exactly the case for not doing the new logic only...?

src/point_one/fusion_engine/parsers/fusion_engine_framer.cc
adamshapiro0: Hmm I suppose. Seems like a wasted instruction though since this is the literal definition of the message size. Normally I would have just moved the size calculation to inside the if statement, but we print it in the warning message so I left it here.

anathan (changes requested):
At a minimum I think we need to have hard bounds checking of the buffer array (I think similar to before), but I also think the conditional logic on arch size is unneeded so I'd rather just hard limit ourselves to 2Gb buffers (2^32 or 31 size)

src/point_one/fusion_engine/parsers/fusion_engine_framer.cc
adamshapiro0: The message size is defined as payload + header, there's no reason for it to ever be set differently. Checking the condition this way using the payload size is guaranteed to include 100% of possible cases, so also doing a check on the message size variable would just be redundant.

adamshapiro0:
Updated to remove that unintended arch condition. The bounds checking should now be 100% covered, since `SetBuffer()` guarantees both minimum and maximum capacity now.

src/point_one/fusion_engine/parsers/fusion_engine_framer.cc
```diff
@@ -189,14 +200,24 @@ p1_ssize_t FusionEngineFramer::OnByte(bool quiet) {
       // Compute the full message size. If the message is too large to fit in
       // the buffer, we cannot parse it. Otherwise, start collecting the
       // message payload.
+      //
+      // Note that while we compute the current_message_size_ here, we
```
anathan, line 204: ok i misunderstood this long comment before. what we're checking for is the case that the header is basically garbage and it's message length is causing a out of bounds condition.

src/point_one/fusion_engine/parsers/fusion_engine_framer.cc
anathan: ok that's fair, the variable names confused me earlier, I see its basically the same check.

anathan (approved):

PR #56, opened 2022-01-17
Improvements/clarifications for working with mixed-content files.

adamshapiro0:


python/examples/raw_message_decode.py
```diff
@@ -0,0 +1,93 @@
...
+
+from examples.message_decode import print_message
+
+
+def decode_message(header, data, offset):
+    # Validate the message length and CRC.
+    if len(data) != header.calcsize() + header.payload_size_bytes:
+        return False
+    else:
+        header.validate_crc(data)
```
anathan, line 21: does this actually do anything if the crc is bad?

python/examples/tcp_client.py
```diff
@@ -38,20 +42,24 @@
 
     # Open the output file if logging was requested.
     if options.output is not None:
-        p1i_path = os.path.splitext(options.output)[0] + '.p1i'
-        if os.path.exists(p1i_path):
-            os.remove(p1i_path)
+        if options.format == 'p1log':
+            p1i_path = os.path.splitext(options.output)[0] + '.p1i'
```
anathan, line 46: its a bit weird that the extension and the input arg are different....perhaps we should support both p1log and p1i?

anathan (approved):
im not sure I understnd the p1i thing, but code looks good.

python/examples/raw_message_decode.py
adamshapiro0: Yes, this function raises an exception on failure.

PR #64, opened 2022-03-08
Added file indexing support to `print_contents` tool.

adamshapiro0:
# New Features
- Added `FileIndex` and `FileIndexBuilder` classes, and added indexing support to `print_contents` tool
- Added option to the decoder to return the byte offset of a message within the incoming byte stream

### DIFF
--- a/python/fusion_engine_client/analysis/file_index.py
+++ b/python/fusion_engine_client/analysis/file_index.py
@@ -61,12 +61,13 @@ class FileIndex(object):
         elif key == 'offset':
             return self._data['offset'] if self._data is not None else None
         else:
-            # Assume the key is a data index or slice descriptor and return a new FileIndex slice.
-            return self.__getitem__(key)
+            raise AttributeError
 
     def __getitem__(self, key):
         if self._data is None:
             return FileIndex()
+        elif isinstance(key, int):
+            return FileIndex(self._data[key:(key + 1)])
         else:
             return FileIndex(self._data[key])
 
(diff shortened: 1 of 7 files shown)


python/fusion_engine_client/analysis/file_index.py
```diff
@@ -0,0 +1,139 @@
+from typing import Union
+
+import os
+
+import numpy as np
+
+from ..messages import MessageType, Timestamp
+
+
+class FileIndex(object):
```
axlan, line 10: This file could use usage comments.

If I understand the high level usage, this is basically a wrapper around a np array of `DTYPE` that serializes to `RAW_DTYPE` for output to a file. Everything else are basically helper functions for getting/building this array.

axlan: I'm a little confused why `RAW_DTYPE` is ever expected to be used outside of this class. I would try to simplify this constructor by not having construction from `RAW_DTYPE` be an option. 

Basically, just simplifying this interface to support the expected usage pattern, which appears to be either:
- Load from file
- Iteratively built from FileIndexBuilder
- Made from a filtered subset of another FileIndex.

Of these it seems like the index_path version of this constructor is really the only one that would be used outside of this file. In that case the logic for construction from a list could live in  FileIndexBuilder for example.

axlan (approved):
Looks good, some comments on possibly making FileIndex a bit simpler to construct.

python/fusion_engine_client/analysis/file_index.py
adamshapiro0: Yeah that makes sense. I removed the `RAW_DTYPE` construction -- that's really an internal detail to save/load -- and made the dtype variables protected. This class still uses `RAW_DTYPE` internally for saving and loading so we can't move it entirely to the builder class. Otherwise this class would have to defer to the builder on every file load. The builder's purpose is really just for building up the data to create an index when one doesn't already exist, but it doesn't deal with any of the raw serialization stuff.

PR #67, opened 2022-03-18
Defined "fast IMU corrections" reset bit.

adamshapiro0:
This distinguishes between IMU corrections that may be observed quickly, and those that take more time to estimate.


axlan (approved):
LGTM

PR #74, opened 2022-04-26
Add bit flags for restarting measurement engine.

axlan:


python/fusion_engine_client/messages/control.py
```diff
@@ -159,6 +159,13 @@ class ResetRequest(MessagePayload):
     RESET_CONFIG = 0x00100000
     ## @}
 
+    ##
+    # @name Restart Hardware Modules
+    # @{
+    ## Restart the measurement engine.
+    RESTART_MEASUREMENT_ENGINE = 0x01000000
```
adamshapiro0, line 166: I think we may want to call this the _GNSS_ measurement engine here and everywhere else to make it clear.

adamshapiro0 (approved):
Looks good, just a suggested name change

PR #75, opened 2022-04-28
Plot displacement over time, and other small analysis tool fixes.

adamshapiro0:
# New Features
- Added a plot of position displacement over time (versus the median position)
- Allow child direct paths as part of the log prefix (`--prefix dir/name` places output in `<log_dir>/dir/name_*.html`)

# Fixes
- Fixed topo plot axes
- Show the time label on all subplots


axlan (approved):
LGTM

PR #76, opened 2022-04-28
Added mixed content support to all example applications.

adamshapiro0:
# New Features
- Use `locate_log()` to all use of mixed-content binary files in all example applications
- Added hostname lookup support to TCP client example applications

# Changes
- Defined a common log base directory variable (defaults to `~/logs` on Linux, or `My Documents/Logs` on Windows), which can also be using the `P1_LOG_BASE_DIR` environment variable

# Fixes
- Remove invalid solutions in `analyze_data.py` example application
- Fixed use of ROS pose message in raw decode example


axlan (approved):
LGTM

PR #78, opened 2022-05-02
Update CalibrationStage definition to insert intermediate stages.

thoshizaki:
# Changes
- Update CalibrationStage definition to insert intermediate stages.

### DIFF
--- a/src/point_one/fusion_engine/messages/solution.h
+++ b/src/point_one/fusion_engine/messages/solution.h
@@ -323,7 +323,16 @@ struct alignas(4) SatelliteInfo {
  */
 enum class CalibrationStage : uint8_t {
   UNKNOWN = 0, ///< Calibration stage not known.
-  MOUNTING_ANGLE = 1, ///< Estimating IMU mounting angles.
+  /**
+   * Performing initial estimation of IMU mounting angles. Reported position
+   * solution may have reduced accuracy.
+   */
+  MOUNTING_ANGLE_INITIAL_CONVERGENCE = 20,
+  /**
+   * Performing final estimation of IMU mounting angles. Position solution
+   * reported with increased accuracy.
+   */
+  MOUNTING_ANGLE_FINAL_CONVERGENCE = 30,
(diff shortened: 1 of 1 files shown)


src/point_one/fusion_engine/messages/solution.h
```diff
@@ -323,7 +323,16 @@ struct alignas(4) SatelliteInfo {
  */
 enum class CalibrationStage : uint8_t {
   UNKNOWN = 0, ///< Calibration stage not known.
-  MOUNTING_ANGLE = 1, ///< Estimating IMU mounting angles.
+  /**
+   * Performing initial estimation of IMU mounting angles. Reported position
+   * solution may have reduced accuracy.
+   */
+  MOUNTING_ANGLE_INITIAL_CONVERGENCE = 20,
```
anathan, line 330: It's unclear to me to if these are the start or end of these stages. Also perhaps the reasons are too directly related to what's happening under the hood, originally we talked.aboit just having a coarse calibration done generically...

adamshapiro0: The stage indicates "this thing is ongoing", and you'd normally be in that stage for some time until moving to the next one.

Normally most of the progression is an internal detail that isn't too important. We figured this one distinction might be useful since the performance may change notably when we get to final convergence, and someone may want to use that information. But we could also skip this for now?

thoshizaki: Shall we make no change at all in this regard for now?

thoshizaki: Under fusion-engine-client.

thoshizaki:
Discussed
- We want to use bit mask for calibration stages.
- Keep using MOUNTING_ANGLE for now: close this PR. 
> For both of internal `MOUNTING_ANGLE_INITIAL_CONVERGENCE` and `MOUNTING_ANGLE_FINAL_CONVERGENCE`, map to `MOUNTING_ANGLE` for external message.

PR #79, opened 2022-05-04
Search for `/logs` directory by default, and fix version string display.

adamshapiro0:
# Changes
- On Linux/Mac, default to `/logs` base directory if it exists, otherwise use `~/logs`

# Fixes
- Fixed missing leading character in version strings in `p1_display` index file


anathan (approved):
looks good, thanks.

PR #82, opened 2022-05-09
Set all bits in `FACTORY_RESET` mask, and other minor cleanup.

adamshapiro0:
# Changes
- Set all 32 bits in `FACTORY_RESET`, including the reserved bits
- Documentation and style cleanup
- Added `IntEnum` wrapper class to eliminate duplicate `str()` overloads.

### DIFF
--- a/python/fusion_engine_client/messages/defs.py
+++ b/python/fusion_engine_client/messages/defs.py
@@ -62,9 +62,9 @@ class Response(IntEnum):
     EXECUTION_FAILURE = 5
     ## The header `payload_size_bytes` is in conflict with the size of the message based on its type or type
     ## specific length fields.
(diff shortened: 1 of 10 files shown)


python/fusion_engine_client/messages/defs.py
```diff
@@ -23,6 +23,9 @@ class SatelliteType(IntEnum):
     SBAS = 8
     IRNSS = 9
 
+    def __str__(self):
```
axlan, line 26: Since this is used in a bunch of places does it make sense to make a common parent class for these?

axlan (approved):
LGTM

python/fusion_engine_client/messages/defs.py
adamshapiro0: Yes, that's probably a good idea.

PR #84, opened 2022-05-16
Removed APPLIED and STEERING from AppliedSpeedType and SteeringType values.

adamshapiro0:


wbrannon (approved):
I agree that these variable names make more sense.

PR #85, opened 2022-05-23
Add flag for rebooting processor

axlan:
Cherry picked from https://github.com/PointOneNav/fusion-engine-client-internal/pull/53


python/fusion_engine_client/messages/control.py
```diff
@@ -163,6 +163,8 @@ class ResetRequest(MessagePayload):
     # @{
     ## Restart the GNSS measurement engine.
     RESTART_GNSS_MEASUREMENT_ENGINE = 0x01000000
+    ## Reboot the FusionEngine processor.
```
adamshapiro0, line 166: These should all say "navigation processor" instead of "FusionEngine processor" for consistency with the param name. Same in the C++.

adamshapiro0 (approved):
Looks good

PR #88, opened 2022-05-27
Fix filter for invalid RelativeENUPositionMessage

axlan:
RelativeENUPositionMessage can have invalid position data base on reference station match so need to actually check data.


alexanderbarton (approved):
Looks good to me.

PR #97, opened 2022-07-14
Various fixes for Python tools.

adamshapiro0:
# Changes
- Added partial name matching for message type names in `print_contents`
- Changed `-m` (message type) and `-t` (time range) definitions in `print_contents` to match `p1_display`

# Fixes
- Fixed missing system time messages when using an index file in `print_contents`
- Fixed message index check in `print_contents`

### DIFF
--- a/python/bin/print_contents.py
+++ b/python/bin/print_contents.py
@@ -100,10 +100,21 @@ other types of data.
         # Convert to lowercase to perform case-insensitive search.
         type_by_name = {k.lower(): v for k, v in message_type_by_name.items()}
         for name in options.type:
-            message_type = type_by_name.get(name.lower(), None)
+            lower_name = name.lower()
+            message_type = type_by_name.get(lower_name, None)
             if message_type is None:
-                print("Unrecognized message type '%s'." % name)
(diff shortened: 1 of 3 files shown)


python/bin/print_contents.py
```diff
@@ -51,14 +53,14 @@ def print_message(header, contents, one_line=False):
         '-s', '--summary', action='store_true',
         help="Print a summary of the messages in the file.")
     parser.add_argument(
-        '--time', type=str, metavar='[START][:END]',
+        '-m', '--message-type', type=str, action='append',
+        help="An optional list of class names corresponding with the message types to be displayed. "
```
axlan, line 57: I may be misremembering how this argparse works, but isn't the usage:
`-m MSG1 -m MSG2` ? Usually when I hear a CLI variable described as a list I assume it means `-m MSG1,MSG2`

axlan (approved):
LGTM

python/bin/print_contents.py
adamshapiro0: Yes that's correct. I may actually just make it support both for convenience.

PR #98, opened 2022-07-18
Switched to mainline gpstime (0.6.2) for Windows leap seconds fixes.

adamshapiro0:
This resolves issues downloading/updating the leap seconds file when necessary, particularly on Windows machines.

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -3,8 +3,6 @@ construct>=2.10.0
 
 # Required for analysis tools only.
 argparse-formatter>=1.4
-# Note: Using the P1 fork of gpstime until Windows and microsecond fixes are mainlined.
-https://github.com/PointOneNav/gpstime/archive/f9e2ab58a8beeeafee992d87a0eafc50887ba849.zip#egg=gpstime
-#gpstime>=0.4.6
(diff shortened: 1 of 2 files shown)


adamshapiro0:
Note: Holding off on merging this temporarily. `gpstime` switched to using `setuptools_scm` in 0.5.0. Unfortunately, `setuptools_scm` requires a complete git repo, so that means it's no longer possible to pull in our fork of `gpstime` using an HTTPS archive. Instead, the user is now required to have `git` installed on their machine.

We can potentially work around this with some complicated hoop jumping for `setuptools_scm` (https://github.com/pypa/setuptools_scm#git-archives), or by reverting back to hard-coding the version number into our fork of `gpstime` and removing `setuptools_scm`, however our remaining changes to `gpstime` have been submitted to the mainline in https://gitlab.com/jrollins/gpstime/-/merge_requests/1. Once they are merged, we can switch to using that through `pip` instead. That's preferable.

adamshapiro0:
Update: Our `gpstime` fixes were mainlined in version 0.6.2 (see https://gitlab.com/jrollins/gpstime/-/merge_requests/1). Switching to version 0.6.2 on PyPI.

PR #99, opened 2022-07-21
Store P1 time in `MeasurementTimestamps`, and add `VehicleTickMeasurement` definition.

adamshapiro0:
# New Features
- Added `VehicleTickMeasurement` definition (message type 11004)

# Changes
- Redefined `MeasurementTimestamps` to explicitly store the computed P1 time value in addition to the input time/time of arrival value
- Improved wheel speed usage documentation


axlan (approved):
LGTM

PR #104, opened 2022-07-25
Fixed str() operators for wheel measurement classes.

adamshapiro0:


axlan (approved):
LGTM, seems like you're having the problem I was having last Friday where some of the checks aren't running.

PR #112, opened 2022-08-25
Added raw byte string decode example app, and fixed other issues.

adamshapiro0:
# New Features
- Added `raw_message_decode.py` byte string decode example (renamed other `raw_*.py` apps to `manual_*.py`)
- Added `str()` operator to `MessageHeader`

# Fixes
- Fixed crash when unpacking a `ResetRequest` message


python/examples/manual_message_decode.py
```diff
@@ -0,0 +1,131 @@
...
+        contents.unpack(buffer=data, offset=offset)
+    else:
+        contents = None
+
+    print_message(header, contents)
+
+    return True
+
+
+decode_message.expected_sequence_number = 0
```
axlan, line 67: Hunh, haven't seen that before. What exactly is `decode_message` referencing here? Is it adding an attribute to the function?

python/examples/manual_message_decode.py
```diff
@@ -0,0 +1,131 @@
...
+            header = MessageHeader()
+            offset = header.unpack(buffer=data)
+        except Exception as e:
+            logger.error('Decode error: %s' % str(e))
+            break
+
+        # Read the message payload and append it to the header.
+        data += f.read(header.payload_size_bytes)
+
+        if not decode_message(header, data, offset):
```
axlan, line 130: Minor thing, but most errors in decode_message (a crc or unpack error) will trigger an exception not return False. It would probably be more illustrative to catch those and return false.

axlan (approved):
LGTM

python/examples/manual_message_decode.py
adamshapiro0: Yep, it's a function attribute. Basically the same as just making this a global, except without the dumb syntax limitation that you can't set a global primitive inside a function so you always have to make the global a list or something.

python/examples/manual_message_decode.py
adamshapiro0: Yeah, good point

PR #115, opened 2022-08-26
Added a mixed binary file reader generator class, and other efficiency improvements.

adamshapiro0:
# New Features
- Added a Python generator class for extracting FusionEngine messages from a mixed binary log file
- Added helper functions to `MessagePayload` for P1/system timestamp extraction and derived type checking
- Added intersection support to `TimeRange` helper class

# Changes
- Refactored `p1_print` to use generator class instead of decoder for efficiency
- Moved `Timestamp` and `MeasurementTimestamps` classes into new `timestamp.py` file to avoid circular dependencies (differs from C++)
- Include an EOF marker in index files for more accurate file size checking with mixed binary files


python/bin/p1_print.py
```diff
@@ -145,179 +150,104 @@ def print_message(header, contents, one_line=False):
...
-            # Otherwise, if we do not have an index file, read the next chunk of data and process all of it.
+                    if system_time_sec < newest_system_time_sec:
+                        _logger.warning('System time restart detected after %s.' %
+                                        system_time_to_str(newest_system_time_sec, is_seconds=True))
+                    last_system_time_sec = max(last_system_time_sec, system_time_sec)
+                    newest_system_time_sec = system_time_sec
+
+            total_messages += 1
+            if header.message_type not in message_stats:
+                message_stats[header.message_type] = {
```
axlan, line 229: Its a bit nicer to use a  defaultdict for this sort of pattern.

python/fusion_engine_client/analysis/file_index.py
```diff
@@ -207,7 +210,26 @@ def load(self, index_path, data_path=None, delete_on_error=True):
                     os.remove(index_path)
                 raise ValueError("Index file empty but data file not 0 length. [size=%d B]" % data_file_size)
 
-            # See if the index is larger than the data file.
+            # Get the last entry in the index. If its message type is INVALID, it's a special marker at the end of the
+            # index file indicating the size of the binary data file when the index was created. If it exists, we can
+            # use it to check if the data file size has changed.
```
axlan, line 215: For clarity, what situation is this check detecting?

python/bin/p1_print.py
adamshapiro0: Sure, good idea.

python/fusion_engine_client/analysis/file_index.py
adamshapiro0: This is looking to see if the size of the `.p1log` file changed at all since the index was originally generated. The most common cause of this is someone re-extracting the contents from an existing mixed-binary file, but hitting Ctrl-C in the middle so the new file is only part of the original file.

python/fusion_engine_client/messages/defs.py
```diff
@@ -402,12 +317,44 @@ def get_type_string(cls):
     def get_version(cls) -> int:
         return cls.MESSAGE_VERSION
 
+    @classmethod
+    def is_instance(cls, obj) -> bool:
```
axlan, line 321: I'm confused what the use of these functions is. Is it so you can call:
`MessagePayload.is_instance(obj)`
instead of 
`isinstance(obj, MessagePayload)` ?

python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -0,0 +1,342 @@
...
+                # End of file.
+                self.total_bytes_read = self.input_file.tell()
+                self.logger.debug('EOF reached.')
+                break
+
+            offset_bytes = self.input_file.tell()
+            self.total_bytes_read = offset_bytes
+            self._print_progress()
+
+            if self.total_bytes_read - self.last_print_bytes > 10e6:
```
axlan, line 131: Isn't this just the repeated contents of `_print_progress()`? Did you mean to remove this?

python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -0,0 +1,342 @@
...
+                header.unpack(data, warn_on_unrecognized=False)
+
+                # Check if the payload is too big.
+                if header.payload_size_bytes > MessageHeader._MAX_EXPECTED_SIZE_BYTES:
+                    raise ValueError('Payload size (%d) too large.' % header.payload_size_bytes)
+
+                # Read and validate the payload.
+                payload_bytes = self.input_file.read(header.payload_size_bytes)
+                read_len += len(payload_bytes)
+                if len(payload_bytes) != header.payload_size_bytes:
```
axlan, line 161: If I understand the two possibilities here are:
1. You reached the end of the file
2. You are near the end of the file, but you got an invalid sync so you are reading an invalid message but don't have enough bytes to validate the CRC so you might miss other valid messages before the end?

If that is the case you're catching it's probably worth having a comment to explain.

python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -0,0 +1,342 @@
...
+                    if self.time_range.in_range_started() and self.index_builder is None:
+                        self.logger.debug("End of time range reached. Finished processing.")
+                        break
+                    else:
+                        self.logger.trace("Message not in time range. Skipping.", depth=1)
+                        continue
+
+                # Construct the result. If we're returning the payload, deserialize the payload.
+                result = []
+                if self.return_header:
```
axlan, line 221: Possibly too late to change this, but I find it a lot easier to consume output from these mutli-optional output return types if they return a dict instead of an array or tuple. The consumers don't need to check the full set of enabled outputs, and don't need to be changed if you change the return values.

python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -0,0 +1,342 @@
...
+
+    def __next__(self):
+        return self.next()
+
+    def __getitem__(self, key):
+        """!
+        @brief Limit the returned messages by type or time.
+
+        @warning
+        This operator modifies this class in-place.
```
axlan, line 297: This seems unexpected enough that I would suggest not using the built in `__getitem__` and to use function name that might suggest this behavior more strongly something like `filter_inplace`.

axlan (approved):
A few comments and questions, but LGTM.

python/fusion_engine_client/messages/defs.py
adamshapiro0: Yeah, there's honestly no great reason for it except for consistency. Ideally we wouldn't have either of these functions, and would just call `isinstance(obj, Foo)` and `issubclass(obj, Foo)`. Turns out though that `isinstance(obj, Foo)` works for any value of `obj` (instance, some other type, `None`, etc.), but `issubclass(obj, Foo)` throws a `TypeError` if `obj` is not a class.

So I made this `is_subclass()` helper strictly for convenience since we're now doing that check in a few places within a larger if/else statement. I didn't want all of those places to have to know how to do the right thing and call `inspect.isclass()` since that was getting ugly. Then once I added `is_subclass()` I figured I should add `is_instance()` for completeness.

I can remove it if you think it's confusing though.

python/fusion_engine_client/parsers/mixed_log_reader.py
adamshapiro0: Yep meant to delete this stuff.

python/fusion_engine_client/parsers/mixed_log_reader.py
adamshapiro0: Yeah good idea, will comment

python/fusion_engine_client/parsers/mixed_log_reader.py
adamshapiro0: Yeah, I tend to agree generally with the caveats that:
1. Technically returning a dict and keying into it by strings is less efficient than indexing into a tuple
2. Returning a tuple is more consistent with a lot of `for a, b, c in generator` types of patterns

Probably not too easy to change this one at this point, but for other similar stuff we can definitely consider using a dict instead.

python/fusion_engine_client/messages/defs.py
axlan: Not really a big deal either way, I'd say remove it since it makes it clearer that the `is_subclass` is meant to fix that specific issue and there's not another reason for making the class members.

python/fusion_engine_client/parsers/mixed_log_reader.py
adamshapiro0: Yeah, good call. Ideally this would be able to return a new reader with the additional filter criteria, but since the reader stores a file object internally and there's no trivial way to copy that, we can't really clone the reader. I'll rename this.

python/fusion_engine_client/messages/defs.py
adamshapiro0: Ok I removed `is_instance()`, and documented `is_subclass()` as follows:
```py
    @classmethod
    def is_subclass(cls, obj) -> bool:
        """!
        @brief Check if an object is a _class_, which is derived from @ref MessagePayload.

        This function calls the built-in `issubclass()` operator to check if a specified object is a class derived from
        @ref MessagePayload. Unlike the operator, which raises a `TypeError` if the object is not a class (e.g., if you
        pass in a `None`, or any other object), this function accepts any type for its argument:
        ```py
        issubclass(None, MessagePayload)        # TypeError
        issubclass(float, MessagePayload)       # False
        issubclass(dict, MessagePayload)        # False
        issubclass(PoseMessage, MessagePayload) # True

        MessagePayload.is_subclass(None)        # False
        MessagePayload.is_subclass(float)       # False
        MessagePayload.is_subclass(dict)        # False
        MessagePayload.is_subclass(PoseMessage) # True
        ```

        Note that this function is specifically meant to check classes, not class _instances_. To test if an object is
        an instance of a class derived from @ref MessagePayload, use the `isinstance()` operator instead:
        ```py
        isinstance(None, MessagePayload)          # False
        isinstance(3.6, MessagePayload)           # False
        isinstance(dict(), MessagePayload)        # False
        isinstance(PoseMessage(), MessagePayload) # True
        ```

        @param obj The object to be tested.
        """
```

PR #118, opened 2022-08-31
Fixed reading issues when all message types are requested (empty set).

adamshapiro0:
# Fixes
- Fixed MixedLogReader handling of empty message type set
- Fixed FileIndex slicing crash if user set is empty


axlan (approved):
LGTM

PR #119, opened 2022-09-01
Bulk message rate configuration support.

axlan:
Support commands to help with ergonomics of setting message rates.

# Changes
1. Remove no longer supported Set/GetOutputInterfaceConfigMessage.
2. Allow wild cards for Set/GetMessageRate.
3. Have MessageRateResponse return an array of responses.
4. Some minor refactoring of how flags work.


src/point_one/fusion_engine/messages/configuration.h
```diff
@@ -1425,7 +1261,11 @@ enum class MessageRate : uint8_t {
    * Output this message at this interval. Not supported for all messages or
    * platforms.
    */
-  INTERVAL_10_S = 12
+  INTERVAL_10_S = 12,
+  /**
+   * Restore this message's rate back to its default value.
+   */
+  DEFAULT = 128
```
adamshapiro0, line 1268: I'd suggest changing this to 255 instead, and in the future if we need extra special values we'll just 254 and down. I don't think we need to have this odd man out in the middle of the range.

adamshapiro0 (approved):
Looks great, just one suggestion.

PR #123, opened 2022-10-06
Improve index generation

axlan:
In extract_fusion_engine_log don't generate an index file for the mixed log, and instead optionally generate one for the extracted file.


adamshapiro0 (approved):
LGTM

PR #125, opened 2022-10-20
Add FromString for DataVersion

axlan:


src/point_one/fusion_engine/messages/data_version.h
```diff
@@ -75,6 +77,8 @@ std::ostream& operator<<(std::ostream& stream, const DataVersion& ver);
 
 std::string ToString(const DataVersion& ver);
 
+DataVersion FromString(const char* str);
```
adamshapiro0, line 80: For convenience, might be nice to have an inline `std::string` overload that calls this one

adamshapiro0 (approved):
Looks good

PR #126, opened 2022-10-27
Added current transport query support, and remapped import/export message IDs.

adamshapiro0:
# New Features
- Added `TransportType::CURRENT` definition to query settings on the current interface

# Changes
- Changed message IDs for import/export and platform storage messages to move out of reserved 20000 range

# Fixes
- Fixed `encode_message.py` example app


python/fusion_engine_client/messages/configuration.py
```diff
@@ -783,18 +785,23 @@ class SetMessageRate(MessagePayload):
     def __init__(self,
                  output_interface: InterfaceID = None,
                  protocol: ProtocolType = ProtocolType.INVALID,
-                 message_id: int = ALL_MESSAGES_ID,
+                 message_id: int = None,
```
axlan, line 788: This is a Python typing violation. `Optional[int]` would be the appropriate type.

axlan: I'm also a little confused why you changed this. Typically, the pattern you're using here is done when providing a mutable object as a default parameter value (initializing a list to None instead of []) which is to avoid accidental mutations of the object, but I don't think that's a concern here.

python/fusion_engine_client/messages/configuration.py
```diff
@@ -833,16 +855,21 @@ class GetMessageRate(MessagePayload):
 
     def __init__(self,
                  output_interface: InterfaceID = None,
-                 protocol: ProtocolType = ProtocolType.INVALID,
-                 request_source = ConfigurationSource.ACTIVE,
-                 message_id: int = ALL_MESSAGES_ID):
+                 protocol: ProtocolType = ProtocolType.ALL,
+                 request_source: ConfigurationSource = ConfigurationSource.ACTIVE,
+                 message_id: int = None):
```
axlan, line 860: This and output_interface should be `Optional`

python/fusion_engine_client/messages/configuration.py
```diff
@@ -873,6 +900,27 @@ class RateResponseEntry(NamedTuple):
     configured_rate: MessageRate = MessageRate.OFF
     effective_rate: MessageRate = MessageRate.OFF
 
+    __parent_str__ = object.__str__
+
+    def __str__(self):
+        enum = None
```
axlan, line 906: Since this logic is repeated it probably makes sense to make a helper function.

axlan (approved):
LGTM

python/fusion_engine_client/messages/configuration.py
adamshapiro0: Makes sense, added a `_get_message_type_string()` function.

PR #128, opened 2022-11-04
Added fault control support.

adamshapiro0:
# New Features
- Added `FaultControlMessage` support
- Added `send_command.py` example application


src/point_one/fusion_engine/messages/fault_control.h
```diff
@@ -0,0 +1,195 @@
...
+ *
+ * See @ref FaultControlMessage.
+ */
+enum class FaultType : uint8_t {
+  /**
+   * Take no action.
+   *
+   * Payload format: none
+   */
+  NONE = 0,
```
axlan, line 38: To be totally clear, sending a `FaultType::COCOM` type fault with `CoComType::NONE` clears the fault.

Does sending a `FaultType::NONE` clear a fault in any case, or does it truly act as a noop?

axlan: On a related note, can for instance a COCOM, and GNSS fault run in parallel?

adamshapiro0: No, it does not clear anything. I could just make it `RESERVED` or something?

Yes, you can do both COCOM and GNSS faults at the same time, though COCOM will stop the GNSS engine from updating so the GNSS disable wouldn't have any effect.

src/point_one/fusion_engine/messages/fault_control.h
```diff
@@ -0,0 +1,195 @@
...
+  /**
+   * Force the device to exhibit a fatal error (intended for factory test
+   * purposes only).
+   *
+   * Unlike @ref FaultType::CRASH, a fatal error will send an error notification
+   * to the user, but will still not send a @ref CommandResponseMessage.
+   *
+   * Payload format: none
+   */
+  FATAL_ERROR = 2,
```
axlan, line 58: For both `FATAL_ERROR` and `CRASH` is it worth going into a bit more detail on the expected behavior of the device along the lines of "Will no longer output position data. If watchdog is enabled the device will reset."

src/point_one/fusion_engine/messages/fault_control.h
axlan: Assuming `NONE` is a noop, It might be worth adding a `CLEAR_FAULTS` type.

src/point_one/fusion_engine/messages/fault_control.h
adamshapiro0: Yeah, good idea. I'll add more details to all of them.

src/point_one/fusion_engine/messages/fault_control.h
adamshapiro0: Yeah `NONE` isn't really useful. I'll just redefine it as `CLEAR_ALL`.

python/examples/send_command.py
```diff
@@ -0,0 +1,164 @@
...
+        logger.setLevel(logging.DEBUG)
+    elif options.verbose == 2:
+        logger.setLevel(logging.DEBUG)
+        logging.getLogger('point_one.fusion_engine.parsers').setLevel(logging.DEBUG)
+    else:
+        logger.setLevel(logging.DEBUG)
+        logging.getLogger('point_one.fusion_engine.parsers').setLevel(logging.TRACE, depth=options.verbose - 2)
+
+    # Specify the message to be sent.
+    message = ResetRequest(reset_mask=ResetRequest.HOT_START)
```
axlan, line 62: I think it might be useful to add a couple other example commands as a comment here to illustrate that this same basic structure is used for config control and whatnot. Eventually, it might be worth pulling some of this into a helper function we can reuse in the Quectel runner.

axlan (approved):
Couple small comments, but LGTM

python/examples/send_command.py
adamshapiro0: Sure, I'll add a few others commented out below this one.

PR #132, opened 2022-11-14
Fix issue when loading indexed file.

axlan:


axlan:
This change is unnecessary for this codebase.

PR #137, opened 2022-12-03
Resolved issues with wheel speed plots.

adamshapiro0:
# Changes
- Plot the navigation engine's forward velocity estimate to compare with wheel speeds
- Improved GPS time support in `Timestamp` class

# Fixes
- Fixed crash when plotting gear/direction on wheel speed plot

### DIFF
--- a/python/fusion_engine_client/analysis/analyzer.py
+++ b/python/fusion_engine_client/analysis/analyzer.py
@@ -751,6 +751,30 @@ class Analyzer(object):
         else:
             speed_type = 'Wheel' if wheel_data is not None else 'Vehicle'
 
+        # If plotting speed data, try to plot the navigation engine's speed estimate for reference.
+        nav_engine_speed_mps = None
+        if type == 'speed':
+            # If we have pose messages _and_ they contain body velocity, we can use that.
+            #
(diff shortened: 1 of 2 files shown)


python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -795,35 +823,66 @@ def _plot_wheel_ticks_or_speeds(self, type):
...
-                                                           self._time_source_to_display_name(vehicle_time_source))
-
-        if wheel_time_source is not None and vehicle_time_source is not None:
-            if wheel_time_source != vehicle_time_source:
+        same_time_source = True
+        if wheel_time_source is None:
+            common_time_source = vehicle_time_source
+        elif vehicle_time_source is None:
+            common_time_source = wheel_time_source
+        else:
```
wbrannon, line 831: Will this do what you expect if both time sources are `None`?

wbrannon (approved):
Looks good.

python/fusion_engine_client/analysis/analyzer.py
adamshapiro0: That should be handled earlier when we load the data. If we don't have either wheel or vehicle measurements, we'll return early and not create the figure.

PR #139, opened 2022-12-03
Improved command-line argument handling in `p1_display.py`.

adamshapiro0:
# New Features
- Added optional syntax for specifying boolean argument values (`--foo`, `--foo=false`, `--no-foo`, etc.)
- Added `--plot` option to `p1_display.py` to support specifying which plots to display

# Changes
- Reorganized `p1_display.py` arguments into groups


anathan (approved):
It's getting.to a point that we should have generic python libs we open source separately

adamshapiro0:
Yeah, totally agree.

PR #140, opened 2022-12-03
Added signal status plot and long log truncation support.

adamshapiro0:
# New Features
- Added GNSS signal status plot (availability and usage info)
- Added `p1_display.py --plot` option to specify which plots to generate
- By default, automatically reduce or skip plots that would be extremely slow for long (multi-hour) logs; disable with `--no-truncate`

# Changes
- Highlight and colorize warnings from `p1_display.py`

# Fixes
- Fixed numpy support for `GNSSSatelliteMessage`

### DIFF
--- a/python/fusion_engine_client/analysis/analyzer.py
+++ b/python/fusion_engine_client/analysis/analyzer.py
@@ -28,7 +28,17 @@ from .file_reader import FileReader
 from ..utils import trace
 from ..utils.argument_parser import ArgumentParser, TriStateBooleanAction, CSVAction
 from ..utils.log import locate_log, DEFAULT_LOG_BASE_DIR
+
+
 _logger = logging.getLogger('point_one.fusion_engine.analysis.analyzer')
+__orig_warning = _logger.warning
+def bold_warning(msg, *args, **kwargs):
+    __orig_warning('\nWARNING: %s\n' % msg, *args, **kwargs)
+_logger.warning = bold_warning
+__orig_error = _logger.error
+def bold_error(msg, *args, **kwargs):
+    __orig_error('\nERROR: %s\n' % msg, *args, **kwargs)
+_logger.error = bold_error
 
 
 SolutionTypeInfo = namedtuple('SolutionTypeInfo', ['name', 'style'])
--- a/python/fusion_engine_client/analysis/analyzer.py
+++ b/python/fusion_engine_client/analysis/analyzer.py
@@ -25,21 +25,12 @@ if __name__ == "__main__" and (__package__ is None or __package__ == ''):
 from ..messages import *
 from .attitude import get_enu_rotation_matrix
 from .file_reader import FileReader
-from ..utils import trace
+from fusion_engine_client.utils.trace import HighlightFormatter
 from ..utils.argument_parser import ArgumentParser, TriStateBooleanAction, CSVAction
 from ..utils.log import locate_log, DEFAULT_LOG_BASE_DIR
 
 
 _logger = logging.getLogger('point_one.fusion_engine.analysis.analyzer')
-__orig_warning = _logger.warning
-def bold_warning(msg, *args, **kwargs):
-    __orig_warning('\nWARNING: %s\n' % msg, *args, **kwargs)
-_logger.warning = bold_warning
-__orig_error = _logger.error
-def bold_error(msg, *args, **kwargs):
-    __orig_error('\nERROR: %s\n' % msg, *args, **kwargs)
-_logger.error = bold_error
-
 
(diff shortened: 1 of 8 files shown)


adamshapiro0:
Example satellite status plot:

image

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -27,7 +27,17 @@
 from ..utils import trace
 from ..utils.argument_parser import ArgumentParser, TriStateBooleanAction, CSVAction
 from ..utils.log import locate_log, DEFAULT_LOG_BASE_DIR
+
+
 _logger = logging.getLogger('point_one.fusion_engine.analysis.analyzer')
+__orig_warning = _logger.warning
+def bold_warning(msg, *args, **kwargs):
```
anathan, line 34: eventually these should probably be in a shared code location (in utils)

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -82,10 +92,13 @@ def _data_to_table(col_titles: List[str], col_values: List[List[Any]]):
...
     logger = _logger
 
+    LONG_LOG_DURATION_SEC = 2 * 3600.0
+    HIGH_MEASUREMENT_RATE_HZ = 40.0
+
     def __init__(self, file: Union[FileReader, str], output_dir: str = None, ignore_index: bool = False,
                  prefix: str = '',
                  time_range: Tuple[Union[float, Timestamp], Union[float, Timestamp]] = None,
-                 absolute_time: bool = False,
+                 absolute_time: bool = False, truncate_long_logs: bool = True,
```
anathan, line 101: this seems strange to be true by default?

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -99,6 +112,7 @@ def __init__(self, file: Union[FileReader, str], output_dir: str = None, ignore_
                Both the start and end values may be set to `None` to read all data.
         @param absolute_time If `True`, interpret the timestamps in `time_range` as absolute P1 times. Otherwise, treat
                them as relative to the first message in the file.
+        @param truncate_long_logs If `True`, reduce or skip certain plots if the log extremely long.
```
anathan, line 115: may be nice to indicate a notion of what "extremely long" means.

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -709,6 +744,135 @@ def _plot_data(name, idx, marker_style=None):
...
+            entry_type = 'Satellite'
+        else:
+            # In practice, the signal status plot can be _VERY_ slow to generate for really long logs (multiple hours)
+            # because plotly doesn't handle figures with lots of traces very efficiently. The legacy satellite status
+            # plot doesn't seem to suffer nearly as much since A) it has fewer elements (# SVs vs # signals), and B) it
+            # only supports at most 2 traces per element since it doesn't convey usage type.
+            if self.truncate_data:
+                _logger.warning('Skipping signal status plot for very long log. Rerun with --truncate=false to '
+                                'generate this plot.')
+                self._add_figure(name=filename, title=f'{figure_title} (Skipped)')
```
anathan, line 779: It'd be nice if the figure says why it was skipped (i.e. due to truncation)

anathan (approved):
minor comments. Overall, I'm honestly not sure if truncation is the right default action. Is decimation much harder?

python/fusion_engine_client/analysis/analyzer.py
adamshapiro0: I can change to false by default but that would conflict with the command line behavior. I intentionally made that true by default since most of the time I expect someone running a long log for the first time is likely not even interested in these slower plots, so it's probably better to print a warning and plot stuff quickly than to sit for a long time waiting for stuff you aren't interested in.

python/fusion_engine_client/analysis/analyzer.py
adamshapiro0: Yeah, this should probably really be a formatter class in our argparse utils file. I'll see if I can do that quickly instead. Definitely agree that we want to make that file shared code eventually.

adamshapiro0:
Decimation isn't necessarily harder, maybe slightly more complicated for some data types, but you risk missing important details.

python/fusion_engine_client/analysis/analyzer.py
adamshapiro0: Scratch that, confused myself. This should be part of our `trace.py` addons for the `logging` library.

adamshapiro0: Added a `HighlightFormatter` helper class to both offset the text like this and colorize it too.

PR #143, opened 2022-12-05
Update README.md

anathan:
fixes the md to match the new location of the python files


adamshapiro0 (approved):
Looks good. I also added `p1_print.py` since that was missing.

PR #144, opened 2022-12-08
Fixed bug involving use of `np.argmax`

wbrannon:
Modifications
-
- Added `numpy_utils.py` file, containing only a `find_first` function for now, meant to replace `np.argmax` for boolean arrays.
- Replaced incorrect use of `np.argmax` with `find_first`


adamshapiro0 (approved):
Looks good

PR #145, opened 2022-12-09
Added backwards compatibility support for earlier versions of colorama.

adamshapiro0:
In particular, the awscli setup.py file currently strictly specifies colorama <0.4.5. Any applications including both awscli and fusion-engine-client will hit a pip dependency conflict if we require >=0.4.6.


wbrannon (approved):
Looks good.

PR #146, opened 2022-12-09
Added sky and C/N0 vs time plots.

adamshapiro0:
# New Features
- Added GNSS sky plot
- Added GNSS C/N0 vs time plot
- Allow wildcards in `--plot` specifiers to show multiple plots

### DIFF
--- a/python/fusion_engine_client/analysis/analyzer.py
+++ b/python/fusion_engine_client/analysis/analyzer.py
@@ -838,6 +838,66 @@ class Analyzer(object):
 
         self._add_figure(name='gnss_skyplot', figure=figure, title='GNSS Sky Plot')
 
+    def plot_cn0(self):
+        # The legacy GNSSSatelliteMessage contains data per satellite, not per signal. The plotted C/N0 values will
+        # reflect the L1 signal, unless L1 is not being tracked.
+        result = self.reader.read(message_types=[GNSSSatelliteMessage], **self.params)
+        data = result[GNSSSatelliteMessage.MESSAGE_TYPE]
+
+        if len(data.p1_time) == 0:
+            self.logger.info('No satellite data available. Skipping C/N0 plot.')
+            return
+
+        # Setup the figure.
+        figure = make_subplots(
+            rows=1, cols=1,  print_grid=False, shared_xaxes=True,
+            subplot_titles=['C/N0 (L1 Only)'])
+
+        figure['layout'].update(showlegend=True, modebar_add=['v1hovermode'])
+        figure['layout']['xaxis1'].update(title="Time (sec)", showticklabels=True)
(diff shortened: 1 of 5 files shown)


python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -742,8 +743,163 @@ def _plot_data(name, idx, marker_style=None):
...
+            rows=1, cols=1,  print_grid=False, shared_xaxes=True,
+            subplot_titles=['C/N0 (L1 Only)'])
+
+        figure['layout'].update(showlegend=True, modebar_add=['v1hovermode'])
+        figure['layout']['xaxis1'].update(title="Time (sec)", showticklabels=True)
+        figure['layout']['yaxis1'].update(title="C/N0 (dB-Hz)")
+
+        # Assign colors to each satellite.
+        data_by_sv = GNSSSatelliteMessage.group_by_sv(data)
+        svs = sorted(list(data_by_sv.keys()))
```
wbrannon, line 862: Does this need to be a sorted list?

python/fusion_engine_client/utils/argument_parser.py
```diff
@@ -151,7 +152,12 @@ def _format(option):
                 elif '=' in option:
                     return None
                 else:
-                    return f"{option}[=<{', '.join(TriStateBooleanAction.POSSIBLE_VALUES)}>]"
+                    nonlocal options_printed
```
wbrannon, line 155: Huh I've never seen this `nonlocal` keyword before - pretty neat.

python/setup.py
```diff
@@ -13,6 +13,7 @@
             'argparse-formatter>=1.4',
             'colorama>=0.4.6',
```
wbrannon, line 14: Make sure to rebase your other PR where you change the `colorama` version here

wbrannon (approved):
Looks good.

python/fusion_engine_client/analysis/analyzer.py
adamshapiro0: Not right here technically, but I want to loop over it in order below so the traces on the plot are listed in order. I could sort it there but figured better to do it here just so `svs` is in order if we want to look at it for debugging.

python/fusion_engine_client/utils/argument_parser.py
adamshapiro0: Yeah it's not a super common thing, and they added it in Python 3. It lets you modify a primitive variable that lives outside the scope of a nested function. Before that in Python 2 you had to do ugly stuff to make your primitive into an object to work around the limitation. Something like:
```py
my_number = [3]
def my_func():
    my_number[0] += 5
```

python/setup.py
adamshapiro0: Yep thanks for the reminder

PR #148, opened 2022-12-20
Fixed index generation crash, and improved mixed binary extraction performance.

adamshapiro0:
# Changes
- Added `max_bytes` option to `MixedLogReader` class
- Removed calls to file `tell()` to improve `MixedLogReader` read performance (2.5x speedup)

# Fixes
- Fixed crash when generating an index for a file starting with a pose message
- Fixed titles of position displacement vs time east and north plots

### DIFF
--- a/python/fusion_engine_client/parsers/mixed_log_reader.py
+++ b/python/fusion_engine_client/parsers/mixed_log_reader.py
@@ -1,6 +1,7 @@
 from datetime import datetime
 import logging
 import os
+import sys
 from typing import Union
 
 from ..analysis import file_index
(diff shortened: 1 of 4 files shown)


python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -32,9 +33,10 @@ def __init__(self, input_file, warn_on_gaps: bool = False, show_progress: bool =
         @param warn_on_gaps If `True`, print warnings if gaps are detected in the FusionEngine message sequence numbers.
         @param show_progress If `True`, print file read progress to the console periodically.
         @param generate_index If `True`, generate an index file if one does not exist for faster reading in the future.
-               See @ref FileIndex for details.
+               See @ref FileIndex for details. Ignored if `max_bytes` is specified.
```
axlan, line 36: I guess the implication is that if you would only want the index for the whole file, but that seems like something that would be better left up the the caller. Is there a implementation reason to make these settings mutually exclusive?

axlan (approved):
LGTM

python/fusion_engine_client/parsers/mixed_log_reader.py
adamshapiro0: I thought it best avoid generating an index file that only covers part of the actual `.p1log` file. When we load an index file, the first thing we do is verify that the `.p1log` file size matches the expected size according to the index. If not, we assume that the index may be incomplete/out of date, and we invalidate it. Otherwise, you can get into a situation where you killed processing before the index finished generating, and then next time you run since the index file exists, we load it thinking it is complete and end up quietly omitting the remainder of the data in the `.p1log` file.

If we wanted to explicitly generate a partial index, the index file would need to record the `max_bytes` limit that was used accidentally trigger this condition. Otherwise, the index file would be invalidated immediately on use so there's no point in creating it.

PR #149, opened 2022-12-28
Refactored `FileReader` and supporting classes to enable direct reading of mixed-binary files.

adamshapiro0:
# New Features
- Added support for mixed-binary files to `FileReader` class, using existing `MixedLogReader` helper class
  - This provides efficient read support without the need to extract FusionEngine data to a separate file when the input file also contains non-FusionEngine content
- Added rewind support to `MixedLogReader` class
- Added `str()` and `==` operators to `TimeRange` class
- Added arithmetic operators to `Timestamp` class
- Added `ExtendedBooleanAction` class, with similar argument support to `TriStateBooleanAction`, but enforcing `True/False`
- Added optional `$` suffix for log search patterns (e.g., `p1_display.py 1234$` matches `1234/` but not `1234abcd/`)

# Changes
- Renamed `FileReader` class to `DataLoader`
- Moved `file_index.py` to `parsers/`
- Changed `TimeRange` end value to _exclusive_, consistent with Python `slice()` syntax
- Redefined `TimeRange` to consider _only_ P1 times when checking time boundaries
  - Messages that do not contain P1 time will be considered in-range only when other P1-containing messages have been tested
- Don't extract FusionEngine content from mixed-binary files by default in `locate_log()` helper function (no longer necessary)
- Renamed `MixedLogReader.filter_in_place()` function (previously `filter_inplace()`)
- Allow `MixedLogReader` index generation enable/disable to be controlled after construction
- Continue reading where we left off if the `MixedLogReader` filter criteria is changed part-way into a file read
- Improved unit testing for `TimeRange`, `MixedLogReader`, and `FileReader`

# Fixes
- Fixed `MixedLogReader` message type filtering if no previous types were specified
- Fixed `FileIndex` P1 time `t0` value if the log does not contain any P1-timestamped messages
- Fixed `FileIndex` attribute query if the index is empty
- Fixed line numbers in trace-level debug prints
- Fixed incorrect `bool` test in `find_first()` helper function causing element 0 to fail unexpectedly
- Fixed missing data copies in `FileIndex` time slicing


python/tests/test_mixed_log_reader.py
```diff
@@ -0,0 +1,405 @@
...
+            (EventNotificationMessage, 2.0),
+            (PoseMessage, 3.0),
+            (PoseMessage, 4.0),
+            (EventNotificationMessage, 4.0),
+        ])
+        return messages
+
+    def _generate_mixed_data_with_binary(self, data_path):
+        messages = self._generate_data(data_path, message_types=[
+            (b'1234', None),
```
axlan, line 78: Since this class uses some custom syncing logic, it might be worth having some of the "binary data" include the FE sync to exercise the logic of starting a sync, then needing to go back to resync after the failure is detecting.

python/fusion_engine_client/analysis/file_reader.py
```diff
@@ -302,336 +300,177 @@ def read(self, message_types: Union[list, tuple, set] = None,
...
-                        break
-            except Exception as e:
-                # Since the CRC passed we know we read the correct number of bytes, so if the decode fails because of
-                # a format version mismatch or something, we can simply skip this message. No need to stop reading the
-                # file.
-                self.logger.warning('Error decoding %s payload @ %d: %s' %
-                                    (header.get_type_string(), message_offset_bytes, repr(e)))
+        # Fast-forward the reader to EOF to print out one last progress update. If we already reached EOF and printed
+        # the update, this should have no effect.
+        self.reader.seek_to_eof()
```
axlan, line 468: Maybe I'm missing something, but is the logic around forcing the EOF just to print out a progress message? It seems like a non-trivial amount of complexity for a log message that doesn't add much context.

axlan (approved):
Some small comments. These reader classes are pretty intense at this point. Not a super clear way to make them easier. I guess part of the problem is that there's sort of a combination of the somewhat separate file IO, filtering, timestamping, and indexing functionality. These can't really be separated since it's most efficient to do them in a single pass.

At some point I wonder if it makes the most sense to try to take the low level indexing and make a (possibly multi-threaded) C++ tool to generate it. This could always run and then the rest of the file reader logic could be built to use an existing index. This would make some things a lot simpler (like finding the last 5 instances of a message), but adds the annoying overhead of a C++ dependency.

For now no problem, but I admit it can be hard to follow what's happening at each level of processing for all the possible combinations of flags.

python/tests/test_mixed_log_reader.py
adamshapiro0: Yeah good point. I'll add that.

python/fusion_engine_client/analysis/file_reader.py
adamshapiro0: Yeah more or less. While I was debugging I found having that last printout was actually pretty useful, especially since we do a lot of reads in sequence in `p1_display`. I suppose I could change this to be a bit more direct and have an explicit print last call instead of actually seeking to eof. The benefit with doing it this way though, is there's no risk of printing 100% and then reading some more data.

adamshapiro0:
Regarding C++, we could maybe do something with cython or ctypes to optimize stuff somewhat or possibly link the C++ library in, rather than doing stuff in pure Python. That would definitely need to be an optional feature though. I don't think want to make compiling and linking in C++ code, or even having a C++ compiler present, a strict requirement on these tools.

I do think we could probably simplify the `FileReader` class as you suggest even without introducing C++, however. Right now, some of the complexity comes from trying to generate the index in the same pass as the first data read, and also supporting reading _without_ an index. We could make indexing a prereq, even if we still generate the index with Python `MixedLogReader` code instead of C++. Then, `FileReader` could assume there is an index and some of the `max_messages` support logic could go away since it would always be done inside `MixedLogReader`, vs right now where that is true only if the index exists. The only negative to that is it would require generating the index during the `FileReader` constructor and `open()`, so that would be quite slow to return. Either that or we delay populating the t0 members until the first read. That would require changes in `p1_display`, etc. though.

Even without more changes though, switching `FileReader` to use `MixedLogReader` under the hood in this PR does at least simplify stuff greatly. The `max_messages` stuff is only complicated now because of the possibility of not having an index, and the complication isn't too terrible anymore, at least not as bad as it was before.

Side note: I think we should maybe rename `FileReader` to disambiguate with `MixedLogReader`. I'm thinking:
1. Rename `FileReader` to `DataLoader` or something? Suggestions?
2. Rename `file_reader.py` too, but maybe leave `file_reader.py` and have an alias class `FileReader` for backwards compatibility? Not sure if that's necessary in practice. Maybe not advisable.
3. Move `file_index.py` into `parsers/` along with `MixedLogReader`.

What do you think?

axlan:
My C++ suggestion was mostly to try to address the reason (the possibly increased processing time) that we don't always want to just run the indexing first as a separate operation.

I think long term I would suggest the following changes:

1. Always use/generate an index.
2. The index has a metadata file with: start/stop times and possibly a file checksum. If those are changed, then the index is regenerated.
3. The whole reader FileReader and all the processing that cares about the content of the messages is done using the index. This lets it process the data with random access, and not needing to constantly check start/end times.

Ideally, we could focus on making the index generation as efficient as possible, but that's a separate task.

As for your 1-3 points at the end:
1. Yeah DataLoader seems reasonable.
2. I'm not a fan of aliasing like this unless there's a strong immediate reason, so I would just rename and move.
3. Sounds good.

PR #150, opened 2023-01-06
Nmea version version sub enums

axlan:
This is to support being able to individually set their rates.

### DIFF
--- a/src/point_one/fusion_engine/messages/configuration.h
+++ b/src/point_one/fusion_engine/messages/configuration.h
@@ -1289,6 +1289,16 @@ enum class MessageRate : uint8_t {
    */
   INTERVAL_10_S = 12,
   /**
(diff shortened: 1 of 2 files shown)


src/point_one/fusion_engine/messages/configuration.h
```diff
@@ -1282,6 +1288,16 @@ enum class MessageRate : uint8_t {
    * platforms.
    */
   INTERVAL_10_S = 12,
+  /**
+   * Output this message at this interval. Not supported for all messages or
```
adamshapiro0, line 1292: This is consistent with the existing ones, but might be a good idea to edit all of these to say more like "Output this message once every 30 seconds." and similar. It's fairly apparent from the enum name, but better to be explicit.

adamshapiro0 (approved):
Looks good

PR #151, opened 2023-01-24
Add REGION_BLACKOUT fault.

axlan:


anathan (approved):

src/point_one/fusion_engine/messages/core.h
```diff
@@ -8,5 +8,6 @@
 #include "point_one/fusion_engine/messages/configuration.h"
 #include "point_one/fusion_engine/messages/control.h"
 #include "point_one/fusion_engine/messages/defs.h"
+#include "point_one/fusion_engine/messages/fault_control.h"
```
adamshapiro0, line 11: Doesn't really hurt, but not sure if this belongs in `core.h` since fault injection isn't really something that's part of normal operation.

adamshapiro0 (approved):
Looks good

PR #152, opened 2023-01-26
Added positional bool argument support, and fixed example apps that use the DataLoader class.

adamshapiro0:
# New Features
- Added support for positional arguments that use tri-state or extended bool actions (`foo false` vs `--foo=false`)

# Fixes
- Fixed example apps that import the `DataLoader` class (formerly `FileReader`)


lukes3315 (approved):
Looks good

PR #157, opened 2023-01-31
Reorganized `trace.py` to avoid modifying the `logging` module.

adamshapiro0:
This is a follow-up to #153 to resolve additional conflicts with 3rd party libraries.


axlan (approved):
I'm not super familiar with this stuff, but I do like this is a bit more explicit in how it's manipulating the logging module.

PR #159, opened 2023-02-06
Updated setup.py to meet requirements and added PyPI support to CI script.

adamshapiro0:
# Changes
- Added missing components from `setup.py`, including querying version from `__init__.py`
- Organized optional requirements "extras" and added install instructions to the README
- Added PyPI upload support to CI script on tagged releases


python/fusion_engine_client/__init__.py
```diff
@@ -0,0 +1,3 @@
+__all__ = ['messages', 'parsers', 'utils']
+__version__ = '1.15.0'
```
axlan, line 2: So this is always the most annoying part of the Python release stuff, but it looks like were going with separately specifying the release tag and the version string passed to the setup.py?

I usually see some sort of file generation in the release process like we use in Nautilus for generating/updating this value.

axlan (approved):
LGTM, made a comment that we may want to avoid separately specifying PyPi version from tag.

README.md
```diff
@@ -185,21 +186,38 @@ bazel run -c opt //message_decode -- message_decode/example_data.p1log
...
+See [Point One FusionEngine Python Client](python) for complete details and usage examples.
 
-1. Install Python 3.4 and Pip
-2. Navigate to the `python/` directory and install the requirements:
+#### Install From PyPI
+
+1. Install Python 3.7 (or later) and pip.
+2. Install the `fusione-engine-client` module _including_ extra requirements used by the analysis and processing tools:
+   ```bash
+   python3 -m pip install fusion-engine-client[all]
```
anathan, line 196: what with the [all] ?

anathan: derp reading further.

python/README.md
```diff
@@ -71,17 +72,54 @@ for the latest API documentation.
     
 ### Usage Instructions
 
-To use the Python library:
+#### Install From PyPI
 
-1. Install Python 3.4 (or later) and pip.
-2. Install the Python requirements:
+1. Install Python 3.7 (or later) and pip.
+2. Install the `fusione-engine-client` module _including_ extra requirements used by the analysis and processing tools:
```
anathan, line 78: While more technically correct, I think this language is confusing. We should just say "it installs the analysis and processing tools and the associated dependencies"

python/fusion_engine_client/__init__.py
anathan: like to derive it from the tagname in the GA action?

python/setup.py
```diff
@@ -1,21 +1,84 @@
...
+        'Intended Audience :: Developers',
+        'Intended Audience :: End Users/Desktop',
+        'Intended Audience :: Science/Research',
+        'License :: OSI Approved :: MIT License',
+        'Operating System :: MacOS :: MacOS X',
+        'Operating System :: Microsoft :: Windows',
+        'Operating System :: POSIX',
+        'Programming Language :: Python',
+        'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.8',
```
anathan, line 57: should this include 3.7?

python/setup.py
```diff
@@ -1,21 +1,84 @@
...
+    description='Point One FusionEngine Library',
+    long_description="""\
+Point One FusionEngine protocol support for real-time interaction and control, plus post-processing data analysis tools.
+
+See https://github.com/PointOneNav/fusion-engine-client for full details. See https://pointonenav.com/docs/
+for the latest FusionEngine message specification.
+""",
+    long_description_content_type='text/markdown',
+    author='Point One Navigation',
+    author_email='support@pointonenavigation.com',
```
anathan, line 44: support@pointonenav.com

anathan (approved):
minor typos to fix but good.

python/fusion_engine_client/__init__.py
adamshapiro0: So I agree - definitely don't want to specify it in a lot of places. I actually modify `setup.py` to read this string from here (borrowed that from pyserial's setup.py file), so this + the git tag are the only two places the version is specified.

@anathan in theory I agree, best would be to just specify it once via the git tag. That way there's no risk of forgetting to update this file before a release. Unfortunately, in order for it to be set as a `__version__` variable within the loaded module at runtime, the build step would need to produce that file dynamically. I couldn't quite figure out how to get the `build` module to do that. Also, it would mean that the version string would not be set at all if anyone was doing active development and not installing a wheel from pip.

So I figured this was the best compromise for now: set the version string here, which automatically informs `setup.py`, and then also specify it in the git tag later. But I'm totally open to other ideas.

Another idea would be to write a tagging script that sets the git tag from the version string instead of the other way around like `git tag -a v$(python setup.py --version)`. Maybe we do that? (Reference: https://stackoverflow.com/questions/6786555/automatic-version-number-both-in-setup-py-setuptools-and-source-code)

adamshapiro0: Ok, I added a new `tag_release.sh` script. @axlan @anathan do you guys mind looking that over real quick?

When we want to tag a release, we can now run `tag_release.sh 1.16.0` or similar, and it will update the Python and C++ version strings/macros (I added a `version.h` for C++), and then create a git tag.

python/setup.py
adamshapiro0: Hah oops

python/setup.py
adamshapiro0: Technically we have it set to require >3.7. We have not tested on anything earlier than 3.8. We can if we want to in the future, but 3.8 was released in 2019 and 3.7 support ended in 2020 so this is probably safe.

python/README.md
adamshapiro0: Ok, changed to simply:
```
2. Install the `fusione-engine-client` module, including all analysis and data processing tools:
```

python/setup.py
adamshapiro0: I updated the README to say 3.8 or later instead of 3.7

scripts/tag_release.sh
```diff
@@ -0,0 +1,199 @@
...
+
+# Check if the new version is older than the current version.
+cat <<EOF | python3
+from packaging import version
+import sys
+
+prev_version = version.parse("${PREV_VERSION}")
+new_version = version.parse("${VERSION}")
+
+if new_version <= prev_version:
```
axlan, line 148: You should add some error message to explain why the script exited here. Something along the lines of:
"This branch was previously tagged with ${PREV_VERSION} which is a newer version number then the specified tag ${VERSION}"

axlan (approved):
LGTM

PR #160, opened 2023-02-07
Renamed heading message from RelativeENUHeading to HeadingMeasurement.

lukes3315:
# Changes
- Renamed heading message from `RelativeENUHeading` to `HeadingMeasurement` and moved to `measurements.h/py`.
- Corrected documentation for heading vector direction.


src/point_one/fusion_engine/messages/defs.h
```diff
@@ -39,7 +39,7 @@ enum class MessageType : uint16_t {
   POSE_AUX = 10003, ///< @ref PoseAuxMessage
   CALIBRATION_STATUS = 10004, ///< @ref CalibrationStatusMessage
   RELATIVE_ENU_POSITION = 10005, ///< @ref RelativeENUPositionMessage
-  RELATIVE_ENU_HEADING = 10006, ///< @ref RelativeENUHeadingMessage
+  HEADING_MEASUREMENT = 10006, ///< @ref HeadingMeasurement
```
adamshapiro0, line 42: Let's move this into the sensor measurement block of values below, so change to 11001 here and in python

adamshapiro0: (and move it down below IMU_MEASUREMENT)

adamshapiro0: You also need to update the `to_string()` function below, including reording to match when you move this value below IMU

src/point_one/fusion_engine/messages/measurements.h
```diff
@@ -440,6 +440,69 @@ struct alignas(4) VehicleTickMeasurement : public MessagePayload {
   uint8_t reserved[3] = {0};
 };
 
+/**
+ * @brief The heading angle (in degrees) with respect to true north,
+ *        pointing from the primary antenna to the secondary antenna.
+ * @ingroup solution_messages
+ *
+ * @note
+ * All data is timestamped using the P1 Time values, which is a monotonic
```
adamshapiro0, line 449: "the P1 Time values" --> "Point One Time"

src/point_one/fusion_engine/messages/defs.h
lukes3315: ok doing now

python/fusion_engine_client/messages/measurements.py
```diff
@@ -509,3 +509,130 @@ def to_numpy(cls, messages):
...
+  Position std (ENU) (m): {self.position_std_enu_m[0]:.2f}, {self.position_std_enu_m[1]:.2f}, {self.position_std_enu_m[2]:.2f}
+  Heading (deg): {self.heading_true_north_deg:.2f}
+  Baseline distance (m): {self.baseline_distance_m:.2f}"""
+
+    @classmethod
+    def calcsize(cls) -> int:
+        return cls._STRUCT.size + MeasurementTimestamps.calcsize()
+
+    @classmethod
+    def to_numpy(cls, messages: Sequence['HeadingMeasurement']):
```
adamshapiro0, line 628: You're missing an import now that you moved this out of `solution.py`, and it's causing your CI to fail. You'll want to add:
```
from typing import Sequence
```
to the top of the file after `import struct`

python/fusion_engine_client/messages/__init__.py
```diff
@@ -10,7 +10,7 @@
     GNSSSatelliteMessage.MESSAGE_TYPE: GNSSSatelliteMessage,
     CalibrationStatus.MESSAGE_TYPE: CalibrationStatus,
     RelativeENUPositionMessage.MESSAGE_TYPE: RelativeENUPositionMessage,
-    RelativeENUHeadingMessage.MESSAGE_TYPE: RelativeENUHeadingMessage,
+    HeadingMeasurement.MESSAGE_TYPE: HeadingMeasurement,
```
adamshapiro0, line 13: Sort below with IMU to match enum order please

adamshapiro0 (approved):
Looks good

PR #161, opened 2023-02-08
Updated CMake flow to support FetchContent, and improved README instructions.

adamshapiro0:
# Changes
- Reorganized `CMakeLists.txt` and renamed CMake options flags to allow fusion-engine-client to be imported into a user project using `FetchContent`
- Added a standalone example project using CMake `FetchContent` to import and use this library
- Added explicit README instructions recommending how to include this library in a bigger project for CMake and Bazel

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@ endif()
 # Define the project and setup the compiler toolchain. This will establish
 # default compiler/linker flags. If the user specifies a cross-compilation
 # toolchain (-DCMAKE_TOOLCHAIN_FILE=...), it will be applied now.
-project(p1_fusion_engine_client)
+project(p1_fusion_engine_client VERSION 1.15.2)
 
 # Set additional compilation flags.
 if (MSVC)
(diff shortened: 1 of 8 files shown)


anathan (approved):
The new cmake syntax is ugly but makes sense.

adamshapiro0:
I don't really think it's _that_ bad, honestly. To me, it's basically no more ugly or complicated than Bazel workspace `git_repository()` syntax.

In any case, if we want to add `find_package()` support as another option in the future for anyone installing the library to their system, we can certainly do that (though for sure our strong recommendation is to _not_ rely on stuff installed on the system). For that, the best option would be to add proper `pkg-config` support, and then use CMake's built-in `FindPkgConfig` module. For example:
```
find_package(PkgConfig)
pkg_check_modules(fusion_engine_client REQUIRED fusion_engine_client>=1.16.0)
```

That way we don't have to write a custom `Find*.cmake` file, and it's also easy to use via `pkg-config` for people using other build systems (autoconf, qtmake, etc.). Shouldn't be _too_ bad to add, just a little bit of work at some point.

I don't think that would work for Linux/Mac developers though, so the current way with `FetchContent` is preferred anyway.

PR #163, opened 2023-02-10
Added a helper function to convert from UTC/GPS time to P1 time.

adamshapiro0:
# New Features
- Added `DataLoader.convert_to_p1_time()` to convert a timestamp or list of timestamps to P1 time

# Fixes
- Corrected the Y2K GPS epoch time


python/fusion_engine_client/analysis/data_loader.py
```diff
@@ -507,6 +510,103 @@ def get_index(self):
...
+            # very long log, we don't read a ton of data from disk when the caller is only interested in a short
+            # snippet. We assume the requested timestamps won't be too far out of the time range specified for the
+            # cached messages.
+            self.logger.debug('Using existing cached pose data for P1 time conversion.')
+            pose_data = self.data[PoseMessage.MESSAGE_TYPE]
+        else:
+            result = self.read(message_types=[PoseMessage], return_numpy=True)
+            pose_data = result[PoseMessage.MESSAGE_TYPE]
+
+        idx = ~np.logical_or(np.isnan(pose_data.p1_time), np.isnan(pose_data.gps_time))
```
wbrannon, line 550: In what scenario would P1 time be NAN while GPS time is non-nan, or vice versa? I'm guessing it has something to do with initialization.

python/fusion_engine_client/analysis/data_loader.py
```diff
@@ -507,6 +510,103 @@ def get_index(self):
...
+        # GPS timestamps or already P1 timestamps. We don't need to convert anything to GPS time, so we can skip right
+        # to the P1 conversion step below.
+        if isinstance(times, np.ndarray) and not assume_utc:
+            return_scalar = False
+            time_sec = times
+        else:
+            # If the user passed in an iterable object (list, tuple numpy array, etc.), convert all entries to GPS or
+            # P1 values in seconds.
+            try:
+                iter(times)
```
wbrannon, line 582: I've never put in a check like this myself to see if an object was iterable, but I guess it turns out that this may be the best way to do it - weird.

python/fusion_engine_client/analysis/data_loader.py
```diff
@@ -2,13 +2,16 @@
...
 from collections import deque
 import copy
-from datetime import datetime
+from datetime import datetime, timezone
 import io
 import os
 
+from gpstime import gpstime, unix2gps
 import numpy as np
+import scipy as sp
```
wbrannon, line 11: FYI: it doesn't seem like there's a universal way to abbreviate "scipy", although I guess `sp` is the most-used abbreviation: https://import-as.github.io/import/scipy.html

It seems like in the scipy documentation, it doesn't really ever use a `import scipy` statement but rather uses `from scipy import ____`. Doesn't matter too much.

wbrannon (approved):
Looks good.

python/fusion_engine_client/analysis/data_loader.py
adamshapiro0: GPS time can be nan if you start up in a place without GPS, like inside a parking garage. Ideally P1 time should never be nan, but it's always good to check just to be safe.

python/fusion_engine_client/analysis/data_loader.py
adamshapiro0: Yeah me neither. Stack overflow said this was the way, which is admittedly surprising but seems to work as expected.

python/fusion_engine_client/analysis/data_loader.py
adamshapiro0: Yeah, for scipy I've most commonly seen either `import scipy as sp` (this way) and `from scipy import x`. In general, for really big libraries like scipy and numpy where you're probably gonna call lots of different functions and classes, I prefer to use `import thing` (or `import really_long_thing as short_alias`), that way everything is scoped under the `thing` name like `thing.ClassIWant` and not just wholesale dumped into your file's namespace. For numpy, `import numpy as np` is extremely common, and for scipy `import scipy as sp` is also pretty common and consistent with numpy, which almost always is present too if you're using scipy.

I find scoping like that especially important in a file within a library that is meant for someone else to import, like this one. There, if someone wanted to do `from data_loader import *`, if we had just done `from numpy import *`, then the caller would _also_ unexpectedly get all numpy stuff suddenly in their file. With this way, they technically _will_ have `np` present in their file, which they didn't ask for, but at least it's just one thing. Still though, I personally wouldn't recommend doing `from data_loader import *` to avoid that problem altogether. I'd suggest just `from data_loader import DataLoader`.

PR #165, opened 2023-02-17
Added missing TRACE definition to trace.py module.

adamshapiro0:
This way, anyone importing trace as logging (instead of importing the Python logging module directly) will have logging.TRACE, similar to logging.DEBUG, etc.


axlan (approved):
LGTM

PR #166, opened 2023-02-21
Fixed issues in heading measurement plot.

adamshapiro0:
# Fixes
- Fixed primary heading calculation
- Fixed timestamps on heading plot
- Fixed 2-sigma envelope if there are any nans in the heading data (heading not available)


lukes3315 (approved):
looks good to me

PR #167, opened 2023-02-23
Suppress expected numpy warning.

axlan:
This is a really minor thing, but this warning is expected in the situation where we're doing this cast, so we should suppress it.


adamshapiro0 (approved):
Looks good

PR #168, opened 2023-02-24
Do not strip line breaks from argument help strings.

adamshapiro0:
Not sure why the default argparse HelpFormatter explicitly separates and indents them correctly but then strips them out.


adamshapiro0:
Nevermind, this change is not necessary. `FlexiHelpFormatter` does this correctly, we just need to prefix list items with `-`, otherwise they get grouped together as a single paragraph and reflowed.

PR #169, opened 2023-02-27
Apply `assume_utc` to `datetime` objects without a timezone.

adamshapiro0:
# Fixes
- Apply `assume_utc` to both `float` values and `datetime` objects where no timezone is specified (`tzinfo`)
- Added `ShutdownRequest` to message type dictionary


wbrannon (approved):
Looks good.

PR #170, opened 2023-03-03
Added GNSS system (constellation) and frequency band enable/disable support.

adamshapiro0:
# New Features
- Added GNSS system (constellation) and frequency band enable/disable support

# Changes
- Moved `SatelliteType` definitions into new `signal_defs.h`, which is included by `defs.h`

# Fixes
- Fixed possible divide by zero when printing reader processing status
- Fixed Python `GetConfigMessage` constructor arguments

### DIFF
--- a/python/fusion_engine_client/messages/signal_defs.py
+++ b/python/fusion_engine_client/messages/signal_defs.py
@@ -1,4 +1,5 @@
 import re
+from typing import List, Union
 
 import numpy as np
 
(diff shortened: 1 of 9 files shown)


python/fusion_engine_client/messages/signal_defs.py
```diff
@@ -34,10 +35,87 @@ class SatelliteType(IntEnum):
...
+    def bit_mask_to_string(cls, mask: int):
+        systems = cls.bit_mask_to_systems(mask)
+        return ', '.join(str(s) for s in systems)
+
+
 class SignalType(IntEnum):
     UNKNOWN = 0
 
 
+class FrequencyBand(IntEnum):
```
axlan, line 79: For basically all the code in FrequencyBandMaskVal and FrequencyBand, you could probably reuse the same logic as SatelliteTypeMaskVal and SatelliteTypeMask by making some sort of class factory. Probably not worth it for now, but if we add a third bitmask we should probably make a generator.

src/point_one/fusion_engine/messages/configuration.h
```diff
@@ -91,6 +91,20 @@ enum class ConfigType : uint16_t {
...
+   * A bitmask indicating which GNSS constellations are enabled.
+   *
+   * Payload format: `uint32_t` (see @ref sat_type_masks)
+   */
+  ENABLED_GNSS_SYSTEMS = 50,
+
+  /**
+   * A bitmask indicating which GNSS frequency bands are enabled.
+   *
+   * Payload format: `uint16_t` (see @ref freq_band_masks)
```
axlan, line 104: Why is this a `uint16_t` while `sat_type_masks` is a `uint32_t`? Seems like it would make more sense to have them be the same type. Ideally, since we generally use 4 byte alignment, I would assume they'd both be `uint32_t`.

axlan (approved):
Minor comments, but LGTM.

python/fusion_engine_client/messages/signal_defs.py
adamshapiro0: Yeah, I thought the same thing. If/when we add another bitmask, we can generalize this a bit to clean things up.

src/point_one/fusion_engine/messages/configuration.h
adamshapiro0: I did `uint16_t` just because realistically there are not a lot of bands so 16 is more than enough, but you're right that there's no need to limit it here and might be better to consider alignment. I'll change to `uint32_t` instead.

PR #174, opened 2023-03-07
Try to print the candidate header/payload in `binary_message_decode.py` example.

adamshapiro0:
# Changes
- Print the payload size and contents if possible when decode fails

# Fixes
- Fixed handling of single-char user byte input (`2E A` vs `2E 0A`)


axlan (approved):
LGTM.

As a side note, I've been thinking about trying to make FE specifications for one of the general binary tools like ImHex or GNU poke if we keep needing to analyze corrupt logs.

PR #175, opened 2023-03-08
[FUS-672] Add flag to SetConfigMessage to revert to default.

axlan:


adamshapiro0 (approved):
Looks good, don't forget the Python

PR #176, opened 2023-03-10
Fixed truncation of non-time-dependent data, and other cleanup.

adamshapiro0:
# Changes
- Added more detail to event notification printouts
- Made tabs vs spaces consistent in message `str()` operators

# Fixes
- Fixed truncation of non-time-dependent data (e.g., thresholds) if the size of the data happens to match the number of time elements


axlan (approved):
LGTM

PR #177, opened 2023-03-10
Fixed binary decode example app and event description packing.

adamshapiro0:
# Changes
- Added a git message template for release tagging script

# Fixes
- Fixed `binary_message_decode.py` example if user string has bytes packed together (2e31 vs 2e 31)
- Fixed event notification packing if description is `str` and not `bytes`


axlan (approved):
LGTM

PR #178, opened 2023-03-10
Added helper functions to make enum printing more user-friendly and robust.

adamshapiro0:
# New Features
- Added `IntEnum` metaclass support for creating new enum instances for unrecognized values
  - This allows us to be more robust when using an older library version with a newer device firmware

### DIFF
--- a/python/fusion_engine_client/messages/defs.py
+++ b/python/fusion_engine_client/messages/defs.py
@@ -111,26 +111,40 @@ class MessageType(IntEnum):
 
     RESERVED = 20000
 
+    @classmethod
+    def get_type_string(cls, type, include_value=True, raise_on_unrecognized=False):
+        return MessageType.static_to_string(type, include_value=include_value,
+                                            raise_on_unrecognized=raise_on_unrecognized)
 
     @classmethod
-    def get_type_string(cls, type):
+    def static_to_string(cls, value, include_value=True, raise_on_unrecognized=False):
         try:
-            if isinstance(type, str):
-                # Convert a string name to a message type (e.g., 'POSE' -> MessageType.POSE).
-                type = MessageType[type.upper()]
-            else:
-                # Convert an int to a MessageType. If `type` is already a MessageType, it'll pass through.
-                type = MessageType(type)
-
-            return '%s (%d)' % (type.name, type.value)
-        except (KeyError, ValueError):
+            return super().static_to_string(value, include_value=include_value, raise_on_unrecognized=True)
+        except (KeyError, ValueError) as e:
+            # For MessageType, if the user specifies an unrecognized value, we return:
+            # - RESERVED - Value defined for internal use only
+            # - UNKNOWN - Value not recognized
+            #
+            # We don't use the default static_to_string() behavior of returning "<Unrecognized>".
+            string_name = None
             try:
-                if int(type) >= MessageType.RESERVED:
-                    return 'RESERVED (%s)' % str(type)
+                int_value = int(value)
+                if value >= MessageType.RESERVED:
+                    string_name = 'RESERVED'
             except BaseException:
-                pass
+                # Value was not an integer. Let it pass through and print its str() representation below.
+                int_value = value
 
-            return 'UNKNOWN (%s)' % str(type)
+            if string_name is None:
+                if raise_on_unrecognized:
+                    raise e
+                else:
+                    string_name = 'UNKNOWN'
+
+            if include_value:
+                return '%s (%s)' % (string_name, str(int_value))
+            else:
+                return string_name
 
 
 class MessageHeader:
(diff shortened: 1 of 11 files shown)


python/fusion_engine_client/messages/defs.py
```diff
@@ -111,26 +111,40 @@ class MessageType(IntEnum):
...
-            if isinstance(type, str):
-                # Convert a string name to a message type (e.g., 'POSE' -> MessageType.POSE).
-                type = MessageType[type.upper()]
-            else:
-                # Convert an int to a MessageType. If `type` is already a MessageType, it'll pass through.
-                type = MessageType(type)
-
-            return '%s (%d)' % (type.name, type.value)
-        except (KeyError, ValueError):
+            return super().static_to_string(value, include_value=include_value, raise_on_unrecognized=True)
```
axlan, line 122: I'm a bit confused what the purpose of this call is. The parent of this class is IntEnum which doesn't define this?

python/fusion_engine_client/messages/defs.py
```diff
@@ -111,26 +111,40 @@ class MessageType(IntEnum):
...
     RESERVED = 20000
 
+    @classmethod
+    def get_type_string(cls, type, include_value=True, raise_on_unrecognized=False):
+        return MessageType.static_to_string(type, include_value=include_value,
+                                            raise_on_unrecognized=raise_on_unrecognized)
 
     @classmethod
-    def get_type_string(cls, type):
+    def static_to_string(cls, value, include_value=True, raise_on_unrecognized=False):
```
axlan, line 120: Is there a reason to make this a static method like this instead of overriding __str__ or __repr__ ?

python/fusion_engine_client/utils/enum_utils.py
```diff
@@ -12,8 +12,28 @@ def __str__(self):
...
 
-    def to_string(self, include_value=False):
+    def to_string(self, include_value=True):
         if include_value:
             return '%s (%d)' % (str(self), int(self))
         else:
             return str(self)
+
+    @classmethod
+    def static_to_string(cls, value, include_value=True, raise_on_unrecognized=False):
```
axlan, line 22: From my previous comments I guess I see how this works now, but I still don't understand the need for the explicitly called classmethod or the call to super. It seems like this would be simpler to just do as __repr__ or as instance methods.
I guess for the main use case it seems simpler to call:
`MessageType(value).display_string()`
or
`repr(MessageType(value))`
depending on how you feel about __repre__.

axlan (approved):
LGTM I think the usage of `get_type_string` is a bit more complicated then it needs to be, but I might be missing something.

axlan (changes requested):
Actually, I guess the reason for all this complexity is to support the case where the value does not map to an enum. I'm a bit concerned that this will push the invalid enum handling up stream where it will fail unexpectedly (like calling `.name` on the response object). This potentially violates the type hinting in the response classes. Parsing should fail, or the enum should be set to INVALID, not to an integer. Alternatively, we could make the IntEnum more complicated to make this more transparent (store the "raw" value, and have a default to fallback to.)

python/fusion_engine_client/utils/enum_utils.py
adamshapiro0: So we do have `str()` support as a class member function (not `repr()`, we just let that do its own thing), and this function ultimately uses that in the `str(self)` calls to get the enum value's name. The point of having this function in addition is that we can gracefully support integer values that are not defined as enums in the version of the code you're working with.

For example, say we add a new `EventType::STUFF = 6` in the C++ on the device, but you're using an older version of the Python code. The previous code would do something like:
```py
# Unpack:
self.event_type = EventType(int_value)

# Later, expecting:
#   Event: STUFF (6)
print('Event: %s' % str(self.event_type))
```
but since `STUFF` isn't defined, that `EventType()` call would actually raise an exception, and the message would fail to parse. Or even if the parsing function is modified to handle the unknown value and set `self.event_type = 6`, the user code might raise an exception trying to print the value out.

Using this function, we will now display `Event: <Unrecognized> (6)` without crashing. That way we can add new definitions without having things hit exceptions unexpectedly in cases where they could do something more graceful. Of course, if you need to do something specific with `EventType.STUFF`, you will need to update your Python requirements to do that regardless.

adamshapiro0:
Ok I refactored this based on our discussion to handle all of the unrecognized stuff via the metaclass. That way if you do `MyEnum(1234)`, it will implicitly extend the enum definition to include 1234 and return a proper `MyEnum` instance instead of `int`. Have another look to see if this is better.

axlan (approved):
LGTM.

Like the python struct IDL, and some of our framing stuff, this might be a good candidate for a general open source library.

PR #180, opened 2023-03-14
Added command/response logging support to `EventNotificationMessage`.

adamshapiro0:
# New Features
- Added new `EventType` enums to capture commands and responses
- Display commands and responses on the `p1_display` events table
- Added `IsCommand()` and `IsResponse()` helper functions
- Added `return_in_order` option to `DataLoader.read()` to return a single list of all message types in the order they were recorded
- Added `return_bytes` option to `DataLoader.read()` to return the original binary header+payload for each message

# Changes
- Made `to_string()` functions `constexpr` where possible (requires C++14; functions set as `inline` for C++11)
- Added consistent `repr()` formatting for all message classes
- Added auto-registration of `MessagePayload` subclasses instead of explicitly managed dict

### DIFF
--- a/python/fusion_engine_client/analysis/analyzer.py
+++ b/python/fusion_engine_client/analysis/analyzer.py
@@ -48,7 +48,13 @@ _SOLUTION_TYPE_MAP = {
 }
 
 
-def _data_to_table(col_titles: List[str], col_values: List[List[Any]]):
+def _data_to_table(col_titles: List[str], values: List[List[Any]], row_major: bool = False):
+    if row_major:
+        # If values is row major (outer index is the table rows), transpose it.
+        col_values = list(map(list, zip(*values)))
+    else:
+        col_values = values
+
     table_html = '<table><tr style="background-color: #a2c4fa">'
     for title in col_titles:
         table_html += f'<th>{title}</th>'
@@ -1577,8 +1583,8 @@ Gold=Float, Green=Integer (Not Fixed), Blue=Integer (Fixed, Float Solution Type)
             return
 
         # Read the data.
-        result = self.reader.read(message_types=[EventNotificationMessage],
-                                  remove_nan_times=False, return_in_order=True, **self.params)
+        result = self.reader.read(message_types={MessageType.EVENT_NOTIFICATION} | COMMAND_MESSAGES | RESPONSE_MESSAGES,
+                                  remove_nan_times=False, return_in_order=True, return_bytes=True, **self.params)
         data = result[None]  # In-order reads return data with the key None.
 
         if len(data.messages) == 0:
@@ -1586,14 +1592,33 @@ Gold=Float, Green=Integer (Not Fixed), Blue=Integer (Fixed, Float Solution Type)
             return
 
         table_columns = ['Relative Time (s)', 'System Time (s)', 'Event', 'Flags', 'Description']
-        table_data = [[], [], [], [], []]
-        table_data[0] = [f'{(m.system_time_ns - self.reader.get_system_t0_ns()) / 1e9:.3f}' for m in data.messages]
-        table_data[1] = [f'{(m.system_time_ns) / 1e9:.3f}' for m in data.messages]
-        table_data[2] = [str(m.event_type) for m in data.messages]
-        table_data[3] = [f'0x{m.event_flags:016X}' for m in data.messages]
-        table_data[4] = [m.event_description.decode('utf-8') for m in data.messages]
+        rows = []
+        system_t0_ns = self.reader.get_system_t0_ns()
+        for message, message_bytes in zip(data.messages, data.messages_bytes):
+            system_time_ns = message.get_system_time_ns()
+            if isinstance(message, EventNotificationMessage):
+                event_type = message.event_type
+                flags = message.event_flags
+                description_str = EventNotificationMessage.event_description_to_string(message.event_type,
+                                                                                       message.event_description)
+            else:
+                flags = None
+                if message.get_type() in COMMAND_MESSAGES:
+                    event_type = EventType.COMMAND
+                else:
+                    event_type = EventType.COMMAND_RESPONSE
+                description_str = EventNotificationMessage.event_description_to_string(message.event_type,
+                                                                                       message_bytes)
 
-        table_html = _data_to_table(table_columns, table_data)
+            rows.append([
+                f'{(system_time_ns - system_t0_ns) / 1e9:.3f}' if system_time_ns is not None else 'N/A',
+                f'{system_time_ns / 1e9:.3f}' if system_time_ns is not None else 'N/A',
+                event_type.to_string(include_value=True),
+                f'0x{flags:016X}' if flags is not None else 'N/A',
+                description_str,
+            ])
+
+        table_html = _data_to_table(table_columns, rows, row_major=True)
         body_html = f"""\
 <h2>Device Event Log</h2>
 <pre>{table_html}</pre>
--- a/python/fusion_engine_client/analysis/analyzer.py
+++ b/python/fusion_engine_client/analysis/analyzer.py
@@ -1577,20 +1577,20 @@ Gold=Float, Green=Integer (Not Fixed), Blue=Integer (Fixed, Float Solution Type)
             return
 
         # Read the data.
-        result = self.reader.read(message_types=[EventNotificationMessage], remove_nan_times=False, **self.params)
-        data = result[EventNotificationMessage.MESSAGE_TYPE]
+        messages = self.reader.read(message_types=[EventNotificationMessage], remove_nan_times=False,
+                                    return_in_order=True, **self.params)
 
-        if len(data.messages) == 0:
+        if len(messages) == 0:
             self.logger.info('No event notification data available.')
             return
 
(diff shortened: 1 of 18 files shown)


adamshapiro0:
Example event table output:

image

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -1577,22 +1590,42 @@ def plot_events(self):
             return
 
         # Read the data.
-        result = self.reader.read(message_types=[EventNotificationMessage], remove_nan_times=False, **self.params)
-        data = result[EventNotificationMessage.MESSAGE_TYPE]
+        result = self.reader.read(message_types={MessageType.EVENT_NOTIFICATION} | COMMAND_MESSAGES | RESPONSE_MESSAGES,
```
axlan, line 1593: Are we expecting to ever have `COMMAND_MESSAGES` in the log? I guess we talked about that for the primary heading device sending data to the secondary, but is that what we'd want to display in this table?

python/fusion_engine_client/messages/control.py
```diff
@@ -430,21 +451,71 @@ def pack(self, buffer: bytes = None, offset: int = 0, return_buffer: bool = True
...
+                        message_repr = repr(message)
+                    except ValueError as e:
+                        pass
+            else:
+                message_repr = '<Malformed>'
+
+            newline = '\n'
+            return f'{message_repr}{newline}' \
+                   f'Data ({len(event_description)} B): {" ".join("%02X" % b for b in event_description)}'
+        elif isinstance(event_description, str):
```
axlan, line 514: This violates your type hinting?

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -1577,22 +1590,42 @@ def plot_events(self):
...
-
-        table_html = _data_to_table(table_columns, table_data)
+        rows = []
+        system_t0_ns = self.reader.get_system_t0_ns()
+        for message, message_bytes in zip(data.messages, data.messages_bytes):
+            system_time_ns = message.get_system_time_ns()
+            if isinstance(message, EventNotificationMessage):
+                event_type = message.event_type
+                flags = message.event_flags
+                description_str = EventNotificationMessage.event_description_to_string(message.event_type,
```
axlan, line 1609: Based on how this is being used, it seems like this would make more sense as a member function? I guess I find the usage here a bit more confusing then:

```python
if isinstance(message, EventNotificationMessage):
                event_type = message.event_type
                flags = message.event_flags
                description_str = message.event_description_to_string()
            else:
                flags = None
                description_str = repr(message)
```

This seems simpler then your current implementation that re-unpacks the known command types.

python/fusion_engine_client/analysis/data_loader.py
```diff
@@ -218,7 +220,7 @@ def generate_index(self, show_progress=False):
                 self.data[MessageType.POSE] = prev_data
 
     def read(self, *args, **kwargs) \
-            -> Dict[MessageType, MessageData]:
+            -> Dict[Optional[MessageType], MessageData]:
```
axlan, line 223: Is this actually a `Union[List[MessageData], Dict[MessageType, MessageData]]` ?

axlan: Ah I see at the very end you add it to a None entry. I guess I'd find that a bit more confusing then the return signature I'm proposing here since `Dict[Optional[MessageType], MessageData]` could have a mix a `None` and not `None` keys. If you're effectively returning a list, you should probably just return a list.

src/point_one/fusion_engine/messages/control.h
```diff
@@ -383,12 +384,35 @@ struct alignas(4) VersionInfoMessage : public MessagePayload {
...
   enum class EventType : uint8_t {
+    /**
+     * Event containing a logged message string from the device.
+     */
     LOG = 0,
+    /**
+     * Event indicating a device reset occurred. The event flags will be set to
+     * the requested reset bitmask, if applicable (see @ref ResetRequest). The
+     * payload will contain a string describing the cause of the reset.
+     */
```
axlan, line 395: I guess if we're logging a command, and a command response, do we need the RESET and CONFIG_CHANGE values. I don't think we're using CONFIG_CHANGE at all currently, and we're not using `RESET` except as a command response.

src/point_one/fusion_engine/messages/configuration.h
```diff
@@ -158,7 +159,7 @@ enum class ConfigType : uint16_t {
  *
  * @return The corresponding string name.
  */
-inline const char* to_string(ConfigType type) {
+P1_CONSTEXPR_FUNC const char* to_string(ConfigType type) {
```
axlan, line 162: Ideally, we'd also make the struct constructors P1_CONSTEXPR_FUNC as well. I've had issues since:
`explicit InterfaceID(TransportType type, uint8_t index = 0)` isn't a `constexpr`.

python/fusion_engine_client/messages/defs.py
```diff
@@ -237,7 +269,7 @@ def pack(self, buffer: bytes = None, offset: int = 0, payload: bytes = None, ret
             return self.calcsize()
 
     def unpack(self, buffer: bytes, offset: int = 0, validate_crc: bool = False,
-               warn_on_unrecognized: bool = True) -> int:
+               warn_on_unrecognized: bool = True, ignore_sync: bool = False) -> int:
```
axlan, line 272: Are you actually using this? It seems a bit redundant to fix the preamble in the event unpack and also add this feature.

axlan (approved):
A few organizational suggestions but overall LGTM.

python/fusion_engine_client/analysis/analyzer.py
adamshapiro0: Typically not. Generally, the device will not echo the command directly back to a diagnostic interface. It will only send a `COMMAND` event notification. So in practice, this shouldn't happen under normal circumstances.

Given that, I figured it might be useful to include it here just in case the log _does_ somehow have commands. Useful for debugging, or if users are doing something custom for logging or generating p1log files.

python/fusion_engine_client/messages/control.py
adamshapiro0: Ah yep

python/fusion_engine_client/analysis/data_loader.py
adamshapiro0: Yes, I originally implemented it that way and returned a list in in-order mode or a dict otherwise. That would be `List[MessagePayload]` though -- a `MessageData` class is a container that holds _all_ of the requested messages. That had consequences for the calling code though when I went to add the ability to return the bytes.

Maybe I'll change to returning the `MessageData` object? That contains two lists: one for payload instances, one for bytes. Basically, return `result[None]` if returning in order.

axlan: Yeah since MessageData is already just a set of lists that makes sense.

src/point_one/fusion_engine/messages/control.h
adamshapiro0: The idea was that `RESET` and `CONFIG_CHANGE` events go to any user with event notifications turned on, so they're always visible for a user to say "where did I do that hot start?", etc. Commands/responses, however, are only sent if diagnostics are enabled.

Worth noting that it doesn't look like `CONFIG_CHANGE` is actually used by the system at the moment. It's defined in the FE library, but not currently implemented in the system.

python/fusion_engine_client/messages/defs.py
adamshapiro0: Yes currently, but not sure we need it anymore. I added this before deciding to do the preamble fixup instead. I can remove it.

src/point_one/fusion_engine/messages/configuration.h
adamshapiro0: Sure I can add that

python/fusion_engine_client/analysis/analyzer.py
adamshapiro0: Yeah might need to think about this a bit. I agree a member function would be simpler. For the `else` case though, I wanted it to use the same format as a command packed within an event (just with some things missing if not applicable like system time). The current version looks like:
```
[GET_CONFIG (13101), source=ACTIVE, type=ENABLED_GNSS_FREQUENCY_BANDS]
Data (28 B): 2E 31 00 00 E9 06 06 E0 02 00 2D 33 00 00 00 00 04 00 00 00 00 00 00 00 33 00 00 00
```

The `repr()` doesn't have the data bytes portion. I suppose I could just implement it here though.

axlan: Why though? Some of the messages are KB large? It seems like you'd only want that for corrupted messages which would never make it here.

adamshapiro0: Corrupted messages will still be printed here, just not the decoded contents. The main use for this is diagnosing issues if a user gets an unexpected response or sends an incorrectly formed command.

At KB sized messages, this will definitely be quite big. I may truncate the printout in that case. That would mainly be for platform storage/export queries though. Right now, we explicitly do not output response events for those to avoid wasting bandwidth since platform storage is automatically output for any interfaces with diagnostic data enabled.

PR #181, opened 2023-03-15
Truncate the byte string displayed for command/response events.

adamshapiro0:
# Changes
- Truncate the displayed byte string to max 128 B

# Fixes
- Fixed incorrect variable name in message rate repr strings
- Replace <> with [] for HTML output to fix missing interface type details
- Removed incorrect `protocol` field from Python `MessageRateResponse` class
- Fixed unexpected config response length/data member variables after unpack
- Handle config responses with no payload gracefully (these can happen on an `UNAVAILABLE` response)
- Fixed config struct default value mismatch between C++ and Python

### DIFF
--- a/python/fusion_engine_client/messages/configuration.py
+++ b/python/fusion_engine_client/messages/configuration.py
@@ -864,10 +864,16 @@ class ConfigResponseMessage(MessagePayload):
 
     def unpack(self, buffer: bytes, offset: int = 0) -> int:
         parsed = self.ConfigResponseMessageConstruct.parse(buffer[offset:])
+
         self.__dict__.update(parsed)
         del self.__dict__['config_length_bytes']
         del self.__dict__['config_data']
-        self.config_object = _conf_gen.CONFIG_MAP[parsed.config_type].parse(parsed.config_data)
+
+        if parsed.config_length_bytes > 0:
+            self.config_object = _conf_gen.CONFIG_MAP[parsed.config_type].parse(parsed.config_data)
+        else:
+            self.config_object = _conf_gen.CONFIG_MAP[parsed.config_type].make_default()
+
         return parsed._io.tell()
 
     def __repr__(self):
--- a/python/fusion_engine_client/messages/configuration.py
+++ b/python/fusion_engine_client/messages/configuration.py
@@ -1095,7 +1095,7 @@ class MessageRateResponse(MessagePayload):
     def __repr__(self):
         result = super().__repr__()[:-1]
         result += f', response={self.response}, interface={self.output_interface}, source={self.config_source}, ' \
-                  f'protocol={self.protocol}]'
+                  f'protocol={self.protocol}, num_entries={len(self.rates)}]'
         return result
 
     def __str__(self):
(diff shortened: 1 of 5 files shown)


python/fusion_engine_client/utils/enum_utils.py
```diff
@@ -7,13 +7,25 @@ class DynamicEnumMeta(EnumMeta):
...
-            return cls[value]
+            name = value
+            try:
+                return cls[name]
+            except KeyError as e:
+                if raise_on_unrecognized:
+                    raise e from None
+                else:
+                    used_values = {int(v) for v in cls}
+                    unused_value = -1
```
axlan, line 23: I find the intent of this a little confusing. This algorithm appears to assume the order of the values is something like:

[ 1 ,10 ,2, 7, -1, 50, -2, 0, -3, ...] where the order of negative values is assumed, but other values can be in whatever order. If that's the case, it seems like this check would be simpler to be:
`unused_value = min(used_values + [0]) - 1` or maybe even clearer as  `unused_value = min(min(a), 0) - 1`

python/fusion_engine_client/messages/configuration.py
```diff
@@ -864,8 +860,16 @@ def pack(self, buffer: bytes = None, offset: int = 0, return_buffer: bool = True
...
         parsed = self.ConfigResponseMessageConstruct.parse(buffer[offset:])
+
         self.__dict__.update(parsed)
-        self.config_object = _conf_gen.CONFIG_MAP[parsed.config_type].parse(parsed.config_data)
+        del self.__dict__['config_length_bytes']
+        del self.__dict__['config_data']
+
+        if parsed.config_length_bytes > 0:
+            self.config_object = _conf_gen.CONFIG_MAP[parsed.config_type].parse(parsed.config_data)
+        else:
```
axlan, line 870: So for this feature where you're supporting default parsing for invalid configuration objects, is the intent that we might send a bad value as part of a `ConfigResponseMessage` and we want to not throw a parsing error? Generally, this would only happen if there's a bug in our firmware or a discrepancy between the C++ and Python interpretation of a ConfigType. I guess I would probably prefer a parsing error to silently filling in default values.

python/fusion_engine_client/utils/enum_utils.py
adamshapiro0: Yeah, I sort of figured negative values are pretty unlikely, but I like your second option better just to be safe.

python/fusion_engine_client/messages/configuration.py
adamshapiro0: I was running into this specifically for `UNAVAILABLE` config responses where we intentionally omit the payload (since we're not handling the request). In that case, we _could_ include the appropriate payload for the requested config type, just with default values, but I think that is actually less clear than just not having a payload since someone might think it's actually populated.

adamshapiro0: Sorry hit enter before I replied to that last sentence.

I tend to agree that having it populate a default struct here in Python is probably equally confusing. I only did this because the Python class doesn't actually have a `config_type` member variable, and just implicitly gets it from `config_object`. So if you tried to print out a `repr()` for an `UNAVAILABLE` message, it would crash.

Maybe a better solution is to add a `__getattr__` handler for `config_type`, and then store the deserialized type internally. I'll think about that a bit.

PR #186, opened 2023-03-22
Read a max of 1 MB when searching for t0.

adamshapiro0:
# Changes
- Added `max_bytes` option to data loader reads
- Read a max of 1 MB when searching for t0 to prevent silently reading for VERY long times on large logs that do not have any messages with P1 and/or system time


axlan (approved):
LGTM

PR #188, opened 2023-03-23
Add serialization logic to not include data for SetConfig when FLAG_REVERT_TO_DEFAULT is set.

axlan:


adamshapiro0 (approved):
Looks good

PR #189, opened 2023-03-23
Repurposed last_differential_time for leap second and corrections info.

adamshapiro0:
# Changes
- Repurposed `GNSSInfoMessage::last_differential_time` field to output:
  - UTC leap second
  - Number of SVs in use
  - GNSS corrections age
  - GNSS corrections baseline distance
- Added a GNSS corrections status plot
- Added `message_version` argument to `MessagePayload.unpack()` API

# Fixes
- Corrected message version number in C++ `HeadingMeasurement` class

### DIFF
--- a/python/fusion_engine_client/messages/solution.py
+++ b/python/fusion_engine_client/messages/solution.py
@@ -237,18 +237,24 @@ class GNSSInfoMessage(MessagePayload):
     @brief Information about the GNSS data used in the @ref PoseMessage with the corresponding timestamp.
     """
     MESSAGE_TYPE = MessageType.GNSS_INFO
-    MESSAGE_VERSION = 0
+    MESSAGE_VERSION = 1
 
+    INVALID_LEAP_SECOND = 0xFF
+    INVALID_AGE = 0xFFFF
+    INVALID_DISTANCE = 0xFFFF
     INVALID_REFERENCE_STATION = 0xFFFFFFFF
 
-    _STRUCT = struct.Struct('<Ifffff')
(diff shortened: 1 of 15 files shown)


python/fusion_engine_client/messages/solution.py
```diff
@@ -287,30 +302,36 @@ def unpack(self, buffer: bytes, offset: int = 0) -> int:
         offset += self.p1_time.unpack(buffer, offset)
         offset += self.gps_time.unpack(buffer, offset)
 
-        offset += self.last_differential_time.unpack(buffer, offset)
-
-        (self.reference_station_id,
+        (self.leap_second, self.num_svs,
```
axlan, line 305: Even though we didn't resize this struct, should we check the message version to avoid setting these values to random garbage, and just invalidate them?

src/point_one/fusion_engine/messages/solution.h
```diff
@@ -204,8 +204,11 @@ struct alignas(4) PoseAuxMessage : public MessagePayload {
  */
 struct alignas(4) GNSSInfoMessage : public MessagePayload {
   static constexpr MessageType MESSAGE_TYPE = MessageType::GNSS_INFO;
-  static constexpr uint8_t MESSAGE_VERSION = 0;
+  static constexpr uint8_t MESSAGE_VERSION = 1;
```
axlan, line 207: I would add a comment that last_differential_time was removed to add these new parameters and that this struct can be used to de-serialize a version 0 message as long as the new parameters are invalidated or not used.

PR #196, opened 2023-03-30
Add interface based configuration commands.

axlan:
Based on https://github.com/PointOneNav/fusion-engine-client-internal/pull/87


adamshapiro0 (approved):
Looks good

PR #197, opened 2023-03-30
Interface config fixes.

axlan:
# Changes
 - Add Python object for interface diagnostic output enabled setting

# Fixes
 - Correct GetConfig version in Python
 - Fix encoding the config data sizes in SetConfig and ConfigResponse


adamshapiro0 (approved):
Looks good

PR #198, opened 2023-04-04
Added new `SystemStatusMessage` definition.

wbrannon:
# New Features
- Defined new `SystemStatusMessage`

### DIFF
--- a/python/fusion_engine_client/messages/device.py
+++ b/python/fusion_engine_client/messages/device.py
@@ -1,14 +1,14 @@
-import struct
+import math
 
-from construct import (Struct, Int64ul, Int16ul, Int8ul, Int16sl, Padding, this, Bytes, PaddedString)
+from construct import (Struct, Int16sl, Padding)
 
-from ..utils.construct_utils import AutoEnum, construct_message_to_string
-from ..utils.enum_utils import IntEnum
+from ..utils.construct_utils import construct_message_to_string
 from .defs import *
+
 
 class SystemStatusMessage(MessagePayload):
     """!
-    @brief System status elements.
+    @brief System status message.
     """
     MESSAGE_TYPE = MessageType.SYSTEM_STATUS
     MESSAGE_VERSION = 0
@@ -18,20 +18,34 @@ class SystemStatusMessage(MessagePayload):
     SystemStatusMessageConstruct = Struct(
         "p1_time" / TimestampConstruct,
         "gnss_temperature" / Int16sl,
-        Padding(131),
+        Padding(118),
     )
 
     def __init__(self):
         self.p1_time = Timestamp()
-        self.gnss_temperature = SystemStatusMessage.INVALID_TEMPERATURE
+        self.gnss_temperature_degc = math.nan
 
     def pack(self, buffer: bytes = None, offset: int = 0, return_buffer: bool = True) -> (bytes, int):
         values = vars(self)
+
+        if math.isnan(values['gnss_temperature_degc']):
+            values['gnss_temperature'] = SystemStatusMessage.INVALID_TEMPERATURE
+        else:
+            values['gnss_temperature'] = int(round(values['gnss_temperature_degc'] * 8))
+        del values['gnss_temperature_degc']
+
         packed_data = self.SystemStatusMessageConstruct.build(values)
         return PackedDataToBuffer(packed_data, buffer, offset, return_buffer)
 
     def unpack(self, buffer: bytes, offset: int = 0, message_version: int = MessagePayload._UNSPECIFIED_VERSION) -> int:
         parsed = self.SystemStatusMessageConstruct.parse(buffer[offset:])
+
+        if parsed['gnss_temperature'] == SystemStatusMessage.INVALID_TEMPERATURE:
+            parsed['gnss_temperature_degc'] = math.nan
+        else:
+            parsed['gnss_temperature_degc'] = parsed['gnss_temperature'] / 8
+        del parsed['gnss_temperature']
+
         self.__dict__.update(parsed)
         return parsed._io.tell()
 
--- a/python/fusion_engine_client/messages/device.py
+++ b/python/fusion_engine_client/messages/device.py
@@ -1,0 +1,47 @@
+import struct
+
+from construct import (Struct, Int64ul, Int16ul, Int8ul, Int16sl, Padding, this, Bytes, PaddedString)
+
+from ..utils.construct_utils import AutoEnum, construct_message_to_string
+from ..utils.enum_utils import IntEnum
+from .defs import *
+
+class SystemStatusMessage(MessagePayload):
+    """!
+    @brief System status elements.
+    """
+    MESSAGE_TYPE = MessageType.SYSTEM_STATUS
+    MESSAGE_VERSION = 0
+
+    INVALID_TEMPERATURE = -32768
+
+    SystemStatusMessageConstruct = Struct(
+        "p1_time" / TimestampConstruct,
+        "gnss_temperature" / Int16sl,
+        Padding(131),
+    )
+
+    def __init__(self):
+        self.p1_time = Timestamp()
+        self.gnss_temperature = SystemStatusMessage.INVALID_TEMPERATURE
+
+    def pack(self, buffer: bytes = None, offset: int = 0, return_buffer: bool = True) -> (bytes, int):
+        values = vars(self)
+        packed_data = self.SystemStatusMessageConstruct.build(values)
+        return PackedDataToBuffer(packed_data, buffer, offset, return_buffer)
+
+    def unpack(self, buffer: bytes, offset: int = 0, message_version: int = MessagePayload._UNSPECIFIED_VERSION) -> int:
+        parsed = self.SystemStatusMessageConstruct.parse(buffer[offset:])
+        self.__dict__.update(parsed)
+        return parsed._io.tell()
+
+    def __repr__(self):
+        result = super().__repr__()[:-1]
+        result += f', gnss_temperature={self.gnss_temperature}]'
+        return result
+
+    def __str__(self):
+        return construct_message_to_string(message=self, title='System Status')
+
+    def calcsize(self) -> int:
+        return self.SystemStatusMessageConstruct.sizeof()
--- a/src/point_one/fusion_engine/messages/defs.h
+++ b/src/point_one/fusion_engine/messages/defs.h
@@ -80,6 +80,8 @@ enum class MessageType : uint16_t {
   GET_MESSAGE_RATE = 13221, ///< @ref GetMessageRate
   MESSAGE_RATE_RESPONSE = 13222, ///< @ref MessageRateResponse
 
+  SYSTEM_STATUS = 13300, ///< @ref SystemStatusMessage
+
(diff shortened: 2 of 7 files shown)


src/point_one/fusion_engine/messages/defs.h
```diff
@@ -80,6 +80,8 @@ enum class MessageType : uint16_t {
   GET_MESSAGE_RATE = 13221, ///< @ref GetMessageRate
   MESSAGE_RATE_RESPONSE = 13222, ///< @ref MessageRateResponse
 
+  SYSTEM_STATUS = 13300, ///< @ref SystemStatusMessage
```
adamshapiro0, line 83: I'd suggest we make group this under 10500+ (below solution messages above) to give it more breathing room, and add a comment to separate device messages like:
```
// Device status messages.
SYSTEM_STATUS = 10500, ///< @ref SystemStatusMessage
```

And matching changes to `defs.py`

src/point_one/fusion_engine/messages/defs.h
```diff
@@ -197,6 +199,10 @@ P1_CONSTEXPR_FUNC const char* to_string(MessageType type) {
     case MessageType::PLATFORM_STORAGE_DATA:
       return "Platform Data Contents";
 
+    // Device-specific status messages.
+    case MessageType::SYSTEM_STATUS:
```
adamshapiro0, line 203: Reminder to order to match 10500 change here too

src/point_one/fusion_engine/messages/device.h
```diff
@@ -0,0 +1,44 @@
...
+namespace fusion_engine {
+namespace messages {
+
+// Enforce 4-byte alignment and packing of all data structures and values.
+// Floating point values are aligned on platforms that require it. This is done
+// with a combination of setting struct attributes, and manual alignment
+// within the definitions. See the "Message Packing" section of the README.
+#pragma pack(push, 1)
+
+struct alignas(4) SystemStatusMessage : public MessagePayload {
```
adamshapiro0, line 21: Make sure to document this class, including brief with ref and version similar to `PoseMessage`, and include an `@ingroup` to add it to a doxygen message group.

I'd suggest we define a new message group `device_status` that falls under `@ingroup messages` see `solution.h` for an example.

src/point_one/fusion_engine/messages/device.h
```diff
@@ -0,0 +1,44 @@
...
+
+  /**
+   * The temperature of the GNSS receiver.
+   *
+   * Stored in units of 1/8 degrees Celsius: `gnss_temperature_degc =
+   * gnss_temperature * 1/8`. Set to `-32768` if invalid.
+   */
+  int16_t gnss_temperature = INVALID_TEMPERATURE;
+
+  uint8_t reserved[131] = {0};
```
adamshapiro0, line 37: Where does 131 come from?

This should probably be 118: 128 B - 2 (`int16_t`) - 8 (`Timestamp`)

python/fusion_engine_client/messages/device.py
```diff
@@ -0,0 +1,47 @@
...
+
+    SystemStatusMessageConstruct = Struct(
+        "p1_time" / TimestampConstruct,
+        "gnss_temperature" / Int16sl,
+        Padding(131),
+    )
+
+    def __init__(self):
+        self.p1_time = Timestamp()
+        self.gnss_temperature = SystemStatusMessage.INVALID_TEMPERATURE
```
adamshapiro0, line 26: In Python, we prefer to store members in standard units (`gnss_temperature_degc`), and do the conversion in the pack and unpack functions.

python/fusion_engine_client/messages/device.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    MESSAGE_TYPE = MessageType.SYSTEM_STATUS
+    MESSAGE_VERSION = 0
+
+    INVALID_TEMPERATURE = -32768
+
+    SystemStatusMessageConstruct = Struct(
+        "p1_time" / TimestampConstruct,
+        "gnss_temperature" / Int16sl,
+        Padding(131),
```
adamshapiro0, line 21: Make sure this matches when you change the C++ reserved size

python/fusion_engine_client/messages/device.py
```diff
@@ -0,0 +1,47 @@
...
+class SystemStatusMessage(MessagePayload):
+    """!
+    @brief System status elements.
+    """
+    MESSAGE_TYPE = MessageType.SYSTEM_STATUS
+    MESSAGE_VERSION = 0
+
+    INVALID_TEMPERATURE = -32768
+
+    SystemStatusMessageConstruct = Struct(
```
adamshapiro0, line 18: Since these construct members are nested within the message classes, I actually don't particularly like the pattern of naming them `FooConstruct` vs simply a generic `Construct`, `SerializationConstruct`, etc. If we use a single name, then it becomes a little easier to maintain the pack/unpack/str code between all the classes since they'll have that in common.

It doesn't come up often, but it has been a bit tedious a few times when we've had to change a pattern across all messages, and we have to change this name for each one.

@axlan thoughts?

python/fusion_engine_client/messages/device.py
```diff
@@ -0,0 +1,47 @@
...
+        "gnss_temperature" / Int16sl,
+        Padding(131),
+    )
+
+    def __init__(self):
+        self.p1_time = Timestamp()
+        self.gnss_temperature = SystemStatusMessage.INVALID_TEMPERATURE
+
+    def pack(self, buffer: bytes = None, offset: int = 0, return_buffer: bool = True) -> (bytes, int):
+        values = vars(self)
```
adamshapiro0, line 29: After this, do conversion to fixed point like:
```py
if math.isnan(values['gnss_temperature_degc']):
    values['gnss_temperature'] = SystemStatusMessage.INVALID_TEMPERATURE
else:
    values['gnss_temperature'] = int(round(values['gnss_temperature_degc'] * 8))
del values['gnss_temperature_degc']
```

python/fusion_engine_client/messages/device.py
```diff
@@ -0,0 +1,47 @@
...
+        packed_data = self.SystemStatusMessageConstruct.build(values)
+        return PackedDataToBuffer(packed_data, buffer, offset, return_buffer)
+
+    def unpack(self, buffer: bytes, offset: int = 0, message_version: int = MessagePayload._UNSPECIFIED_VERSION) -> int:
+        parsed = self.SystemStatusMessageConstruct.parse(buffer[offset:])
```
adamshapiro0, line 34: Similarly, here convert back from fixed point:
```py
if parsed['gnss_temperature'] == SystemStatusMessage.INVALID_TEMPERATURE:
    parsed['gnss_temperature_degc'] = math.nan
else:
    parsed['gnss_temperature_degc'] = parsed['gnss_temperature'] / 8
del parsed['gnss_temperature']
```

python/fusion_engine_client/messages/device.py
```diff
@@ -0,0 +1,47 @@
+import struct
```
adamshapiro0, line 1: This is not used since you're using construct

python/fusion_engine_client/messages/device.py
```diff
@@ -0,0 +1,47 @@
...
+
+from construct import (Struct, Int64ul, Int16ul, Int8ul, Int16sl, Padding, this, Bytes, PaddedString)
+
+from ..utils.construct_utils import AutoEnum, construct_message_to_string
+from ..utils.enum_utils import IntEnum
```
adamshapiro0, line 6: Same here - you're not using enums so don't need `IntEnum` or `AutoEnum`

python/fusion_engine_client/messages/device.py
```diff
@@ -0,0 +1,47 @@
...
+from .defs import *
+
+class SystemStatusMessage(MessagePayload):
```
adamshapiro0, line 9: 2 blank lines before each class definition per PEP8

python/fusion_engine_client/messages/device.py
```diff
@@ -0,0 +1,47 @@
...
+    """!
+    @brief System status elements.
```
adamshapiro0, line 11: System status message

adamshapiro0 (changes requested):
Looks good overall. Let's make some changes and do another pass.

python/fusion_engine_client/messages/device.py
axlan: Just as a general note, PyLint can show you which imports are being used. In VSCode it will grey them out to make it easy to prune https://stackoverflow.com/questions/53357649/detect-unused-imports-in-visual-studio-code-for-python-3 .

src/point_one/fusion_engine/messages/device.h
wbrannon: A couple things here: 
1. I thought we wanted 128 _reserved_ bytes, not 128B for the total size
2. So you don't include the `static constexpr` values in this calculation? I got a value of `131` by including those values in the calculation, and then setting it up to where we had >= 128 reserved bytes

python/fusion_engine_client/messages/device.py
axlan: Since the messages have grown somewhat organically, it probably wouldn't hurt to take a step back at some point and try to rethink the inheritance design. In that sense it probably makes sense to refactor this, but it's outside the scope of this PR.

src/point_one/fusion_engine/messages/device.h
adamshapiro0: 1. @axlan what do you think?
   - @wbrannon to be clear though, the message size must be a multiple of 4B so if we're going for ~128B reserved it either needs to be 126 or 130 so we account for the extra 2 we need since `gnss_temperature` is only 2B.
2. No, the `static` means it's a class-level variable -- a compile time constant in this case since it's `constexpr`. It's not part of an individual instance of the class, so it doesn't count towards the memory used by a message.

axlan (approved):
Functionality looks good, nothing beyond the comments Adam made.

src/point_one/fusion_engine/messages/device.h
```diff
@@ -0,0 +1,64 @@
...
+ * @defgroup device_status System Status Message Definitions
+ * @brief Output messages containing device-specific status information.
+ * @ingroup messages
+ *
+ * See also @ref messages.
+ */
+
+/**
+ * @brief System status: GNSS receiver temperature (@ref
+ *        MessageType::SYSTEM_STATUS, version 0.0).
```
adamshapiro0, line 31: 1.0 (the `MESSAGE_VERSION` constant reflects the minor number)

src/point_one/fusion_engine/messages/device.h
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * @defgroup device_status System Status Message Definitions
+ * @brief Output messages containing device-specific status information.
+ * @ingroup messages
+ *
+ * See also @ref messages.
+ */
+
+/**
+ * @brief System status: GNSS receiver temperature (@ref
```
adamshapiro0, line 30: I think just "System status information" is enough. The receiver temperature is just one detail, no need to list out all of the contents (even though for now there's just one).

PR #199, opened 2023-04-06
Update System Status message string

wbrannon:
`SystemStatusMessage` string before change (using `p1_print`): 
```
System Status [sequence=4, size=152 B, offset=472 B (0x1d8)]
  _io: <_io.BytesIO object at 0x7f49047a0fe0>
  gnss_temperature_degc: 70.0
  p1_time: P1: 2239.646 sec
```

After change: 
```
System Status Message @ P1: 2239.646 sec [sequence=4, size=152 B, offset=472 B (0x1d8)]
  GNSS Temperature: 70.0 deg C
```

### DIFF
--- a/python/fusion_engine_client/messages/device.py
+++ b/python/fusion_engine_client/messages/device.py
@@ -47,6 +47,7 @@ class SystemStatusMessage(MessagePayload):
         del parsed['gnss_temperature']
 
         self.__dict__.update(parsed)
+        del self.__dict__['_io']
         return parsed._io.tell()
 
     def __repr__(self):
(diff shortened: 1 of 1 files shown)


python/fusion_engine_client/messages/device.py
```diff
@@ -55,7 +55,9 @@ def __repr__(self):
         return result
 
     def __str__(self):
-        return construct_message_to_string(message=self, title='System Status')
+        string = 'System Status Message @ %s\n' % str(self.p1_time)
+        string += f'  GNSS Temperature: %.1f deg C\n' % self.gnss_temperature_degc
```
adamshapiro0, line 59: Don't include a trailing \n for the end of the str output

adamshapiro0 (approved):

PR #200, opened 2023-04-07
Redefined measurement messages to separate input/output and use fixed point for raw measurements.

adamshapiro0:
# New Features
- Added `SensorDataSource` enum definition
- Added message definitions for raw (uncorrected) IMU and wheel/vehicle speed

# Changes
- Redefined IMU and wheel messages to separate inputs and outputs (breaking change)
  - Inputs and (new) raw measurement outputs now use fixed-point representation
- Deprecated and renamed `WheelSpeedMeasurement` and `VehicleSpeedMeasurement` version 1.0 (breaking change)
  - These messages were previously used for _both_ input and output, which is no longer supported
- Renamed `MeasurementTimestamps` struct to `MeasurementDetails`


src/point_one/fusion_engine/messages/measurements.h
```diff
@@ -188,6 +275,51 @@ struct alignas(4) IMUMeasurement : public MessagePayload {
   double gyro_std_rps[3] = {NAN, NAN, NAN};
 };
 
+/**
+ * @brief Raw (uncorrected) IMU sensor measurement output (@ref
+          MessageType::RAW_IMU_OUTPUT, version 1.0).
```
axlan, line 280: Missmatch between comment version and message version.

axlan: Oh I see I guess we've been calling version 0, 1.0? I guess we're consistent with that, though not with what to call MESSAGE_VERSION = 1. We usually refer to that as 1.1, but it's wrong about half the time in RelativeENUPositionMessage, CalibrationStatusMessage, and GetConfigMessage .

python/fusion_engine_client/utils/construct_utils.py
```diff
@@ -6,6 +7,31 @@
...
+        if self.invalid is not None and is_signed:
+            max_negative_mag = 2 ** (self.subcon.sizeof() * 8 - 1)
+            if self.invalid == max_negative_mag:
+                self.invalid = -self.invalid
+
+    def _decode(self, obj, context, path):
+        if obj == self.invalid:
+            return math.nan
+        else:
+            return obj * self.scale
```
axlan, line 26: Small thing, but I guess this might have some slightly weird behavior since this will sometimes return an int, and sometimes a float depending on the scale. Probably worth casting to a float for consistency.

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -1131,11 +1145,14 @@ def _plot_wheel_ticks_or_speeds(self, type):
...
-                dt_sec = data.measurement_time[1] - data.measurement_time[0]
-            else:
+
+            if wheel_measurement_type is not None:
+                result = self.reader.read(message_types=wheel_measurement_type, remove_nan_times=False, **params)
+                data = result[wheel_measurement_type.MESSAGE_TYPE]
+                if len(data.measurement_time) == 2:
+                    dt_sec = data.measurement_time[1] - data.measurement_time[0]
+
+            if dt_sec is None and vehicle_measurement_type is not None:
```
axlan, line 1155: So I understand, it seems like `dt_sec` is just use to check whether this plot should be suppressed (The warning it generates seems to be misleading since the file size isn't used as a precondition). It would seem more consistent to also calculate it for  vehicle_measurement_type of both are present and take the max?

python/fusion_engine_client/messages/measurement_details.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    Construct = Struct(
+        "measurement_time" / TimestampConstruct,
+        "measurement_time_source" / AutoEnum(Int8ul, SystemTimeSource),
+        "data_source" / AutoEnum(Int8ul, SensorDataSource),
+        Padding(2),
+        "p1_time" / TimestampConstruct,
+    )
+
+    _STRUCT = struct.Struct('<BB2x')
```
axlan, line 58: I guess it doesn't hurt anything, but is there a specific purpose for having both the construct and struct implementation of the parsing?

axlan: I see it's used inside the other messages that are specified with struct. I suppose you could still use the construct there, but I guess it's probably easier to just clean this up another day.

src/point_one/fusion_engine/messages/measurements.h
adamshapiro0: Yeah basically the system is "this struct is 1.x and the version field is the x". If we want to go to 2.x it'll be a new struct. We haven't actually had to do that yet so not sure how it'll look if we do.

If the constant doesn't match up with the documentation, that's a mistake and we should fix it.

python/fusion_engine_client/utils/construct_utils.py
adamshapiro0: Makes sense

python/fusion_engine_client/messages/measurement_details.py
adamshapiro0: Yeah my thoughts exactly.

python/fusion_engine_client/analysis/analyzer.py
adamshapiro0: Ah yeah this should be conditioned on `self.long_log_detected`.

It does respect the `--truncate=false` -- this stuff is all nested under an `if self.truncate_data:`

I don't think we really need the max. In practice, logs should only have wheel or vehicle speed. We support both just on the VERY rare chance that both are present.

PR #203, opened 2023-04-17
Updated the to_string method for ConfigType to include missing configs.

ericschorling:
# Fixes
- Updated the `to_string()` method for `ConfigType` to include missing configs


wbrannon (approved):
Looks good.

PR #204, opened 2023-04-19
Add FusionEngine message for turning off or modifying atmospheric delay models

wbrannon:
Modifications
-
- Added `AtmosphericDelayModelConfig`, which contains parameters for the ionospheric model (`iono_delay_model`) and the tropospheric model (`tropo_delay_model`)

### DIFF
--- a/python/fusion_engine_client/messages/configuration.py
+++ b/python/fusion_engine_client/messages/configuration.py
@@ -570,9 +570,7 @@ class _ConfigClassGenerator:
 
     class AtmosphericDelayModelConfig(NamedTuple):
         """!
-        @brief Atmospheric delay model configuration settings.
-
-        See @ref VehicleTickMeasurement.
+        @brief Ionospheric and tropospheric delay model configuration.
         """
 
         ## If not OFF -- the ionospheric delay model to be used.
--- a/python/fusion_engine_client/messages/configuration.py
+++ b/python/fusion_engine_client/messages/configuration.py
@@ -33,6 +33,7 @@ class ConfigType(IntEnum):
     ENABLED_GNSS_FREQUENCY_BANDS = 51
     LEAP_SECOND = 52
     GPS_WEEK_ROLLOVER = 53
+    ATMOSPHERIC_DELAY_MODEL = 54
     INTERFACE_CONFIG = 200
     UART1_BAUD = 256
     UART2_BAUD = 257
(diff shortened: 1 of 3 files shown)


axlan (approved):
LGTM

python/fusion_engine_client/messages/configuration.py
```diff
@@ -560,6 +568,24 @@ class HardwareTickConfig(NamedTuple):
         "wheel_ticks_to_m" / Float32l,
     )
 
+    class AtmosphericDelayModelConfig(NamedTuple):
+        """!
+        @brief Atmospheric delay model configuration settings.
+
+        See @ref VehicleTickMeasurement.
```
ryancdougherty, line 575: Why would we look at VehicleTickMeasurement?

python/fusion_engine_client/messages/configuration.py
```diff
@@ -207,6 +208,13 @@ class NmeaMessageType(IntEnum):
     PQTMVER_SUB = 1204
     PQTMTXT = 1205
 
+class IonoDelayModel(IntEnum):
+    OFF = 0
+    KLOBUCHAR = 1
```
ryancdougherty, line 213: Maybe we should add `DEFAULT` to this list. That way we get behaviors like "Off", "On", or "something specific". Maybe I am just saying that to register the thought. In the interest of getting a point release out today, let's not make that change now.

ryancdougherty (approved):
I had one comment about adding `DEFAULT` states to the enums, which would be useful for users resetting things to "normal" if they ever changed them. But for now, I think this is okay.

python/fusion_engine_client/messages/configuration.py
wbrannon: Oops darn copy and paste.

python/fusion_engine_client/messages/configuration.py
axlan: We should probably not add default to config options to specific values. We have a planned implementation for "not set" and there's already a way to set a config option to default generally.

wbrannon: So are we planning to _not_ add a default option here then? The way I'll set it to work, if the user doesn't set it at all, then it'll default to using the models we currently use.

axlan: We should not at a default option here. @wbrannon I'm not sure what you mean by "if the user doesn't set it at all". Typically we handle defaults in the default user config json files, so I assume that's what you're talking about?

wbrannon: @axlan yes the way you described it is what I mean. Thanks for clarifying.

PR #214, opened 2023-05-05
Fixed broken pipe handling in `p1_print`.

adamshapiro0:
This prevents a flood of errors when `p1_print` is piped to another program that exits early (e.g., `p1_print | less`, and then hitting 'q').


axlan (approved):
LGTM

PR #215, opened 2023-05-10
Small plotting and documentation fixes.

adamshapiro0:
# Fixes
- Enable plotly hovermode buttons on all plots
- Allow for a small tolerance in `p1_print` time restart check to account for latency between measurement data and solution output
- Clarified ROS IMU message covariance definition


axlan (approved):
LGTM

PR #216, opened 2023-05-10
Resolved log search issues when specifying local directories/files.

adamshapiro0:
# Changes
- Refactored log search to check: exact file matches, local files/directories, local glob patterns, log dir pattern match

# Fixes
- Fixed incorrect start time printout if `p1_display` uses a different t0 to establish relative time offset

### DIFF
--- a/python/fusion_engine_client/utils/log.py
+++ b/python/fusion_engine_client/utils/log.py
@@ -425,7 +425,7 @@ def locate_log(input_path, log_base_dir=DEFAULT_LOG_BASE_DIR, return_output_dir=
         try:
             result = find_log_file(input_path, candidate_files=CANDIDATE_MIXED_FILES, log_base_dir=log_base_dir,
                                    return_output_dir=return_output_dir, return_log_id=return_log_id,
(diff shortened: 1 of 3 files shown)


python/fusion_engine_client/utils/log.py
```diff
@@ -179,6 +187,8 @@ def find_log_file(input_path, candidate_files=None, return_output_dir=False, ret
     @param log_base_dir The base directory to be searched when performing a pattern match for a log directory.
     @param check_exact_match If `True`, check if `input_path` is the path to a data file. Otherwise, skip this check and
            only perform a pattern search.
+    @param check_pattern_match If `True` and `input_path` does not refer to a log file or directory, perform a pattern
```
axlan, line 190: This interface seems like it would be simpler if it assumed there might be a fuzzy match, and selected the exact match if there was an ambiguity. It seems like there isn't a strong use case for having the `check_exact_match`, and `check_pattern_match` arguments.

axlan (approved):
LGTM

PR #220, opened 2023-06-06
Add a new configuration type for the user device id.

axlan:
# New Features
 - Add USER_DEVICE_ID configuration type.

### DIFF
--- a/src/point_one/fusion_engine/messages/configuration.h
+++ b/src/point_one/fusion_engine/messages/configuration.h
@@ -238,7 +238,9 @@ enum class ConfigType : uint16_t {
   /**
    * A string for identifying a device.
    *
-   * Payload format: `char[24]`
+   * This is a string of ASCII characters padded to 32 bytes with `NULL`.
+   *
+   * Payload format: `char[32]`
    */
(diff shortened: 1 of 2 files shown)


src/point_one/fusion_engine/messages/configuration.h
```diff
@@ -234,6 +234,13 @@ enum class ConfigType : uint16_t {
    * Payload format: `bool`
    */
   ENABLE_WATCHDOG_TIMER = 300,
+
+  /**
+   * A string for identifying a device.
+   *
+   * Payload format: `char[24]`
```
adamshapiro0, line 241: Is this string required to be null terminated? I assume not and that it infers length based on the message size, but probably should document one way or the other

adamshapiro0 (approved):
Looks good

src/point_one/fusion_engine/messages/configuration.h
adamshapiro0: We might want to consider increasing this limit, or eliminating altogether since the protocol itself doesn't really care about an upper bound. It might make sense to have this be platform-specific.

PR #221, opened 2023-06-08
Print UTC time in PoseMessage and GNSSInfoMessage string printout.

adamshapiro0:
# Changes
- Include UTC time in `PoseMessage` and `GNSSInfoMessage` `str()/repr()` operators


axlan (approved):
LGTM

PR #222, opened 2023-06-09
Skip sync byte validation by default in `MessageHeader.unpack()`.

adamshapiro0:
This is typically not needed since the data is almost always read by a decoder, which already performs sync.


axlan (approved):
LGTM

PR #226, opened 2023-06-13
Fixed data indexing and time range issues.

adamshapiro0:
# Changes
- Added `MixedLogReader.clear_filters()` helper function
- Set the log ID to the parent directory when loading from `.` so plot results are stored in `plot_fusion_engine` consistently

# Fixes
- Fixed handling of `NAN` timestamps when applying a time range to a file index
- Fixed time range filtering when reading a message type that does not have P1 time (data will be filtered based on other message types that do have P1 time)
- Fixed incorrect empty response when requesting a list of unavailable message types


wbrannon (approved):
Looks good.

PR #228, opened 2023-06-16
Add L-Band frame extractor.

axlan:
# New Features
 - Add tool for dumping L-Band frame data to a file.

### DIFF
--- a/python/bin/p1_lband_extract
+++ b/python/bin/p1_lband_extract
@@ -1,0 +1,147 @@
+#!/usr/bin/env python3
+
+from fusion_engine_client.utils.trace import HighlightFormatter, BrokenPipeStreamHandler
+from fusion_engine_client.utils.time_range import TimeRange
+from fusion_engine_client.utils.log import locate_log, DEFAULT_LOG_BASE_DIR
+from fusion_engine_client.utils.argument_parser import ArgumentParser, ExtendedBooleanAction
+from fusion_engine_client.utils import trace as logging
+from fusion_engine_client.parsers import MixedLogReader
+from fusion_engine_client.messages import *
+import os
+import sys
+
+# Add the Python root directory (fusion-engine-client/python/) to the import search path to enable FusionEngine imports
+# if this application is being run directly out of the repository and is not installed as a pip package.
+root_dir = os.path.normpath(os.path.join(os.path.dirname(__file__), '..'))
+sys.path.insert(0, root_dir)
+
+root_dir = os.path.normpath(os.path.join(os.path.dirname(__file__), '..'))
+sys.path.append(root_dir)
+
+
+_logger = logging.getLogger('point_one.fusion_engine.applications.lband_extract')
+
+if __name__ == "__main__":
+    parser = ArgumentParser(description="""\
+Extract L-Band corrections data from a log containing FusionEngine L-Band frame messages.
+""")
+
+    parser.add_argument(
+        '--absolute-time', '--abs', action=ExtendedBooleanAction,
+        help="Interpret the timestamps in --time as absolute P1 times. Otherwise, treat them as relative to the first "
+             "message in the file. Ignored if --time contains a type specifier.")
+    parser.add_argument(
+        '-t', '--time', type=str, metavar='[START][:END][:{rel,abs}]',
+        help="The desired time range to be analyzed. Both start and end may be omitted to read from beginning or to "
+             "the end of the file. By default, timestamps are treated as relative to the first message in the file, "
+             "unless an 'abs' type is specified or --absolute-time is set.")
+    parser.add_argument('-v', '--verbose', action='count', default=0,
+                        help="Print verbose/trace debugging messages.")
+    output_parser = parser.add_argument_group('Output Control')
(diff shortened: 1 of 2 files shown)


python/bin/p1_lband_extract
```diff
@@ -0,0 +1,147 @@
...
+import os
+import sys
+
+# Add the Python root directory (fusion-engine-client/python/) to the import search path to enable FusionEngine imports
+# if this application is being run directly out of the repository and is not installed as a pip package.
+root_dir = os.path.normpath(os.path.join(os.path.dirname(__file__), '..'))
+sys.path.insert(0, root_dir)
+
+root_dir = os.path.normpath(os.path.join(os.path.dirname(__file__), '..'))
+sys.path.append(root_dir)
```
adamshapiro0, line 19: Huh this is in p1_print, so I assume it was just copy/pasted here, but not sure why we have both prepend (insert) and append? Also we're doing the same root_dir statement twice. I think the append was just accidental leftovers after https://github.com/PointOneNav/fusion-engine-client/commit/57ddee92272bbae015203332c4ec32c662068748?

python/bin/p1_lband_extract
```diff
@@ -0,0 +1,147 @@
...
+
+    # Read through the data file, searching for valid LBandFrame messages to extract and store in
+    # 'output_dir/<prefix>.p1log'.
+    if options.prefix is not None:
+        prefix = options.prefix
+    elif log_id is not None:
+        prefix = 'lband'
+    else:
+        prefix = os.path.splitext(os.path.basename(input_path))[0]
+    output_path = os.path.join(output_dir, prefix + '.rtcm')
```
adamshapiro0, line 106: L-band data is not guaranteed to be RTCM, so we probably should not use this file extension. Could call it `.lband` but that's a bit weird. If we use a generic extension like `.bin` or something, we might want to call the file `lband.bin` instead of `<prefix>.bin` since that would be too generic.

Or we just have the user specify the output filename at the command line.

python/bin/p1_lband_extract
```diff
@@ -0,0 +1,147 @@
...
+                    fd.write(message.data_payload)
+        except (BrokenPipeError, KeyboardInterrupt) as e:
+            sys.exit(1)
+
+    # Print the data summary.
+    if total_messages > 0:
+        _logger.info('Input file: %s' % input_path)
+        _logger.info('Log ID: %s' % log_id)
+        _logger.info('Total Messages: %d' % total_messages)
+        _logger.info('L-Band Bytes Decided: %d' % bytes_decoded)
```
adamshapiro0, line 145: typo: decoded

python/bin/p1_lband_extract
```diff
@@ -0,0 +1,147 @@
...
+                if isinstance(message, LBandFrameMessage):
+                    bytes_decoded += len(message.data_payload)
+                    total_messages += 1
+                    fd.write(message.data_payload)
+        except (BrokenPipeError, KeyboardInterrupt) as e:
+            sys.exit(1)
+
+    # Print the data summary.
+    if total_messages > 0:
+        _logger.info('Input file: %s' % input_path)
```
adamshapiro0, line 142: Capital File, or lowercase the other prints for consistency

python/bin/p1_lband_extract
```diff
@@ -0,0 +1,147 @@
...
+        prefix = os.path.splitext(os.path.basename(input_path))[0]
+    output_path = os.path.join(output_dir, prefix + '.rtcm')
+
+    _logger.info("Processing input file '%s'." % input_path)
+
+    # Parse the time range.
+    time_range = TimeRange.parse(options.time, absolute=options.absolute_time)
+
+    if options.time is not None:
+        _logger.info('Non-P1 time messages requested and time range specified. Disabling index file.')
```
adamshapiro0, line 114: Since we know we're only requesting a very specific message, we can probably be more clear here and say like "L-band messages do not use P1 time. Disabling index file when applying time range to system timestamps."

That said, I actually wonder in this case if it would be more efficient to _use_ the index and read in all of the messages, and then truncate after the fact using the system time numpy array? Depends on the input file size and time range for sure, but in some cases that might be way faster than totally disabling the index and having to parse the file in slow mode. Same might apply generally in `p1_print` and the other tools. We might want to think about that?

python/bin/p1_lband_extract
```diff
@@ -0,0 +1,147 @@
...
+    if reader.generating_index() and (len(message_types) > 0 or options.time is not None):
+        _logger.info('Generating index file - processing complete file. This may take some time.')
+
+    total_messages = 0
+    bytes_decoded = 0
+
+    with open(output_path, 'wb') as fd:
+        try:
+            for header, message, data, offset_bytes in reader:
+                if isinstance(message, LBandFrameMessage):
```
adamshapiro0, line 133: This is inefficient - it reads in the whole file. We should apply this as a filter to the reader before starting the loop. Then, when there's an index file, it will only read the L-band messages.

adamshapiro0 (approved):
Looks good, but we should think about some of the notes on efficiency a little bit, and also about the output file name.

PR #230, opened 2023-06-21
[FUS-1268] Add portibility defintions for armcc.

axlan:
This is a series of changes to allow the fusion engine client code to be built in armcc.

Some quirks of armcc:
- `NAN` macro bug
- no `alignsas` support
- no `std::function` support
- not `#include <typetraits>`
- No support for glibc file operations.
- `#include <ostream>` brings in fopen.
- `#include <iomanip>` brings in fopen.
- No smart pointers
- `std::vector` brings in fopen
- Doesn’t like `uint8_t test[0];`

### DIFF
--- a/src/point_one/fusion_engine/common/portability.h
+++ b/src/point_one/fusion_engine/common/portability.h
@@ -35,14 +35,58 @@
 #  endif
 #endif
 
-// ssize_t is a POSIX extension and is not supported on Windows.
-#if defined(_WIN32)
+// Support ARM CC quirks
+#ifdef __CC_ARM
+#  include "typedefs.h"
+// ARM CC bug http://www.keil.com/forum/60227/
+#  define __ESCAPE__(__x) (__x)
+#  define P1_HAVE_STD_FUNCTION 0
+#  define P1_HAVE_STD_OSTREAM 0
+#  define P1_HAVE_STD_SMART_PTR 0
+#  define P1_NO_LOGGING 1
+#endif
+
+// Different compilers support different ways of specifying default struct
+// alignment. Since there's no universal method, a macro is used instead.
+#ifdef __CC_ARM
+#  define P1_ALIGNAS(N) __attribute__((aligned(N)))
+#else
+#  define P1_ALIGNAS(N) alignas(N)
+#endif
+
+// ssize_t is a POSIX extension and is not supported on Windows/ARM CC.
+#if defined(_WIN32) || defined(__CC_ARM)
 typedef int32_t p1_ssize_t;
 #elif defined(_MSC_VER)
 typedef int64_t p1_ssize_t;
 #else
 #  include <sys/types.h> // For ssize_t
 typedef ssize_t p1_ssize_t;
+#endif
+
+#ifndef P1_HAVE_STD_OSTREAM
+#  define P1_HAVE_STD_OSTREAM 1
+#endif
+#if P1_HAVE_STD_OSTREAM
+#  include <ostream>
+using p1_ostream = std::ostream;
+#else
+class p1_ostream {
+ public:
+  p1_ostream() = default;
+};
+template <class T>
+inline p1_ostream& operator<<(p1_ostream& stream, const T&) {
+  return stream;
+}
+#endif
+
+#ifndef P1_HAVE_STD_FUNCTION
+#  define P1_HAVE_STD_FUNCTION 1
+#endif
+
+#ifndef P1_HAVE_STD_SMART_PTR
+#  define P1_HAVE_STD_SMART_PTR 1
 #endif
 
 // Support for multi-statement constexpr functions was not added until C++14.
--- a/src/point_one/fusion_engine/messages/data_version.cc
+++ b/src/point_one/fusion_engine/messages/data_version.cc
@@ -7,7 +7,11 @@ namespace fusion_engine {
 namespace messages {
 
 p1_ostream& operator<<(p1_ostream& stream, const DataVersion& ver) {
-  return stream << ToString(ver);
+  if (ver.IsValid()) {
+    return stream << (int)ver.major_version << "." << ver.minor_version;
+  } else {
+    return stream << "<invalid>";
+  }
 }
 
 std::string ToString(const DataVersion& ver) {
--- a/src/point_one/fusion_engine/messages/data_version.cc
+++ b/src/point_one/fusion_engine/messages/data_version.cc
@@ -6,18 +6,17 @@ namespace point_one {
 namespace fusion_engine {
 namespace messages {
 
-std::ostream& operator<<(std::ostream& stream, const DataVersion& ver) {
-  if (ver.IsValid()) {
-    return stream << (int)ver.major_version << "." << ver.minor_version;
-  } else {
-    return stream << "<invalid>";
-  }
+p1_ostream& operator<<(p1_ostream& stream, const DataVersion& ver) {
+  return stream << ToString(ver);
 }
 
 std::string ToString(const DataVersion& ver) {
-  std::stringstream ss;
-  ss << ver;
-  return ss.str();
+  if (ver.IsValid()) {
+    return std::to_string(ver.major_version) + "." +
+           std::to_string(ver.minor_version);
+  } else {
+    return "<invalid>";
+  }
 }
 
 DataVersion FromString(const char* str) {
--- a/src/point_one/fusion_engine/messages/gnss_corrections.h
+++ b/src/point_one/fusion_engine/messages/gnss_corrections.h
@@ -29,7 +29,7 @@ namespace messages {
  * @brief L-Band frame contents (@ref MessageType::LBAND_FRAME, version 1.0).
  * @ingroup gnss_corrections
  */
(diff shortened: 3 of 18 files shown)


src/point_one/fusion_engine/messages/gnss_corrections.h
```diff
@@ -57,9 +57,9 @@ struct alignas(4) LBandFrameMessage : public MessagePayload {
   float doppler_hz = 0;
 
   /**
-   * A pointer to the beginning of the demodulated L-Band frame data.
+   * The beginning of the demodulated L-Band frame data.
    */
-  uint8_t data_payload[0];
+  // uint8_t data_payload[0];
```
anathan, line 62: Is this change intentional?

src/point_one/fusion_engine/common/portability.h
```diff
@@ -40,11 +40,21 @@
 typedef int32_t p1_ssize_t;
 #elif defined(_MSC_VER)
 typedef int64_t p1_ssize_t;
+// Definitions for armcc compiler.
+#elif defined(__MST__)
+#  include "typedefs.h"
+typedef int32_t p1_ssize_t;
 #else
```
anathan, line 47: I'd hope the arm.compiler actually had an arch depending ssizet...crazy that it doesn't.

anathan (approved):
Check that one commented code block please

src/point_one/fusion_engine/messages/gnss_corrections.h
axlan: It's intentional. This has come up before, but technically, zero size struct members aren't allowed in the by the C++ standard. They happen to be supported by GCC and clang. There's no ISO standard way to do this sort of thing unfortunately. It's convenient to have, but I've given up trying to support it for cross compilers.

axlan:
@adamshapiro0 The CI tests are failing for the Documentation build with:

`/home/runner/work/fusion-engine-client/fusion-engine-client/src/point_one/fusion_engine/messages/configuration.h:581: error: Found ';' while parsing initializer list! (doxygen could be confused by a macro call without semicolon) (warning treated as error, aborting now)`

This doesn't make sense to me though since, that `;` is at the end of a struct definition, and nothing has changed around that struct to cause this error. My guess is that my macro change is somehow screwing up the doxy compiler.

Not sure if you have any insight?

src/point_one/fusion_engine/common/portability.h
```diff
@@ -35,8 +35,27 @@
...
+#  define P1_HAVE_STD_SMART_PTR 0
+#  define P1_NO_LOGGING 1
+#endif
+
+// Different compilers support different ways of specifying default struct
+// alignment. Since there's no universal method, a macro is used instead.
+#ifdef __CC_ARM
+#  define P1_ALIGNAS(N) __attribute__((aligned(N)))
+#else
+#  define P1_ALIGNAS(N) alignas(N)
```
adamshapiro0, line 54: Just curious, but regular GCC supports `__attribute__((aligned(N)))` in addition to `alignas(N)`. Is there any benefit to using `alignas`?

src/point_one/fusion_engine/common/portability.h
```diff
@@ -45,6 +64,31 @@ typedef int64_t p1_ssize_t;
...
+#if P1_HAVE_STD_OSTREAM
+#  include <ostream>
+using p1_ostream = std::ostream;
+#else
+class p1_ostream {
+ public:
+  p1_ostream() = default;
+};
+template <class T>
+inline p1_ostream& operator<<(p1_ostream& stream, const T&) {
```
adamshapiro0, line 79: Does `<ostream>` define a default templated no-op operator like this? I would think you wouldn't want to have one, and it would be better to get a compiler error if you try to stream something where << isn't defined?

src/point_one/fusion_engine/messages/data_version.cc
```diff
@@ -6,18 +6,17 @@ namespace point_one {
...
+p1_ostream& operator<<(p1_ostream& stream, const DataVersion& ver) {
+  return stream << ToString(ver);
 }
 
 std::string ToString(const DataVersion& ver) {
-  std::stringstream ss;
-  ss << ver;
-  return ss.str();
+  if (ver.IsValid()) {
+    return std::to_string(ver.major_version) + "." +
```
adamshapiro0, line 15: I'd suggest switching this change back. Generally, if `ToString()` returns a `std::string` I find it's better to have `ToString()` call `<<` instead of the other way around. That way, if someone wants to stream out the object directly `s << foo` they can without allocating heap for a `std::string`.

The exception to that is if the `ToString()` returns `constexpr const char*` like for enums. In that case, `operator<<()` should call `ToString()`.

src/point_one/fusion_engine/messages/gnss_corrections.h
adamshapiro0: We have had an internal ticket to remove these for a while now for similar reasons, so I agree this is the right change.

src/point_one/fusion_engine/common/portability.h
adamshapiro0: Oh I think nevermind - this is effectively a cheap way to define no-ops for all the primitive `operator<<()` overloads that are defined in `basic_stream` normally.

src/point_one/fusion_engine/parsers/fusion_engine_framer.cc
```diff
@@ -18,24 +21,75 @@ using namespace point_one::fusion_engine::messages;
...
-    return stream;
   }
-};
+#endif
+  return stream;
+}
+
+/**
+ * @brief Wrap an integer so it will be output to a stream as its hex
+ * representation.
```
adamshapiro0, line 66: Indent to align with "Wrap"

src/point_one/fusion_engine/parsers/fusion_engine_framer.cc
```diff
@@ -18,24 +21,75 @@ using namespace point_one::fusion_engine::messages;
...
+ *
+ * For example:
+ *
+ * ```cpp
+ *   std::cout << PrintableValue((int16_t)-255) << std::endl;
```
adamshapiro0, line 71: Don't indent the code in the block

adamshapiro0: Also theses should say `HexPrintableValue`

src/point_one/fusion_engine/parsers/fusion_engine_framer.cc
```diff
@@ -70,12 +124,18 @@ void FusionEngineFramer::SetBuffer(void* buffer, size_t capacity_bytes) {
...
+#if P1_HAVE_STD_SMART_PTR
   if (buffer == nullptr) {
     managed_buffer_.reset(new uint8_t[capacity_bytes]);
     buffer = managed_buffer_.get();
   } else if (buffer != managed_buffer_.get()) {
     managed_buffer_.reset(nullptr);
   }
+#else
+  if (buffer == nullptr) {
+    LOG(FATAL) << "P1_HAVE_STD_SMART_PTR required for dynamic memory.";
```
adamshapiro0, line 136: In the no smart pointer case, we could just add an extra bool member var saying that we own the memory, and then `delete[]` the raw pointer in this class's destructor (and also the else if above). The smart pointer was just a convenience to get that same behavior. Honestly we could also just remove it in favor of the bool?

adamshapiro0 (approved):
Looks good. Suggest removing the smart pointer stuff entirely since it's not really needed.

src/point_one/fusion_engine/common/portability.h
axlan: We originally switched to `alignas` since it worked for MVCC, Clang, and GCC. I think it was MVCC that doesn't support `__attribute__((aligned(N)))`. I figure leaving the "most popular" version as the default makes sense.

axlan: At some point we should byte the bullet and define a macro for "structs that have correct padding to align their members, but use 4 byte alignment overall, and don't add implicit padding". If we're using a custom macro already it's not that much more complicated to have:

```
P1_START_PACKED_STRUCT Foo{

} P1_END_PACKED_STRUCT
```

or something similar to actually do this right and get compile time errors if you mess it up.

src/point_one/fusion_engine/messages/data_version.cc
axlan: I understand why the original is preferred, I switched it to this so this function could work regardless of `P1_HAVE_STD_OSTREAM` and to avoid needing to add `P1_HAVE_STD_OSTREAM` checks. I guess the "best" thing is to have both.

src/point_one/fusion_engine/parsers/fusion_engine_framer.cc
axlan: The reason that smart pointers are needed is for all the implicit copy/move constructors. Getting the behavior correct without implicit RAII objects is always a pain. I wanted to avoid it since it adds a lot of additional code, but I guess a compromise would be to delete the implicit constructors.

PR #234, opened 2023-07-05
Fixed issues with UTC and system time extraction.

adamshapiro0:
# Fixes
- Fixed UTC time access in GNSSInfoMessage `str()` operator
- Fixed system time units for timestamped-on-reception measurements
- Mark both GPS and UTC times as approximate on `p1_display` index if applicable


wbrannon (approved):
Looks good.

PR #235, opened 2023-07-07
Added Python 3.6 and 3.7 compatibility.

adamshapiro0:
# Changes
- Lowered `scipy` and `packaging` requirements to versions available for Python 3.6
  - These older versions should be fine for the few features we use from these modules
- Removed the user of `__future__.annotations` (added in Python 3.7) and switched to old-style quoted type hints for self-references


anathan (approved):
lgtm

axlan (approved):
LGTM

PR #239, opened 2023-07-25
Add plotting of dilution of precision

wbrannon:


python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -1063,6 +1063,45 @@ def plot_gnss_signal_status(self):
...
+        figure['layout']['xaxis'].update(title=self.p1_time_label)
+
+        dops = [data.gdop, data.pdop, data.hdop, data.vdop]
+        names = ['GDOP', 'PDOP', 'HDOP', 'VDOP']
+
+        # Assign colors to each DOP type.
+        color_by_dop = self._assign_colors(names)
+
+        # Plot each DOP type.
+        for i in range(len(dops)):
```
ryancdougherty, line 1092: We can do this more pythonically by defining `dops` as a list of tuples of name and data, e.g.
```dops = [('GDOP', data.gdop), ...]```
Then in the loop, something like
```for name, dop in dops:```
I may have that syntax wrong -- this is the kind of thing I always do by finding an example. 

And now you don't need to do any indexing. As a weak, general rule, if you can easily avoid computing indexes, you probably should.

**Review by ryancdougherty — APPROVED** [review]
Looks okay to me. Let's get buy-in from @adamshapiro0 as I have very little experience in this repo.

File: python/fusion_engine_client/analysis/analyzer.py [hunk 2]
```diff
@@ -1063,6 +1063,45 @@ def plot_gnss_signal_status(self):
 
         self._add_figure(name=filename, figure=figure, title=figure_title)
 
+    def plot_dop(self):
+        """!
+        @brief Plot dilution of precision (DOP). This includes geometric, position, horizontal, and vertical DOP.
```
adamshapiro0, line 1068: Minor note: doxygen briefs can only be one sentence. The second sentence becomes part of the "details" section, even if you put it on this line in the code. Generally, this formatting should always be separated:
```
@brief Plot dilution of precision (DOP).

This includes geometric, position, horizontal, and vertical DOP.
```

python/fusion_engine_client/analysis/analyzer.py
adamshapiro0: Agreed. The indexing is fine, but Ryan's suggestion is maybe slightly more pythonic (and the syntax is correct).

However, the `_assign_colors()` function is going to make a dict mapping "element in list" to a color. If you made the `dops` list the way Ryan suggests and passed that to `_assign_colors()`, the keys in the color dict would be the tuples. Then, when you for loop as `name, dop`, you'd have to reconstruct the `(name, dop)` tuple to query the color dict.

Ways around that would be either to not expand the tuple in the for loop:
```
for entry in dops:
    name, dop = entry
    ...
    color_by_dop[entry]
```
or do a list comprehension to assign colors by name:
```
color_by_dop = self._assign_colors([entry[0] for entry in dops])
```

IMO, I'd say either your existing index option or the list comp option are better than keying by tuple. Personally, I might have done the list comp just because when I have to have two independent lists with matching entries, it usually feels to me like it should be one list. But your version works fine too. If this function was more complicated and you had to access those a lot, I'd maybe suggest otherwise.

A _third_ option, which keeps your existing definition but eliminates the explicit indexing, would be to zip the two lists together:
```
for name, dop in zip(names, dops):
```

That's probably slightly cleaner than the indexing loop.

adamshapiro0 (approved):
Looks good

PR #240, opened 2023-08-04
Add L-band decoding example.

axlan:


examples/lband_decode/lband_decode.cc
```diff
@@ -0,0 +1,99 @@
...
+  static constexpr size_t FRAMER_BUFFER_BYTES = 1024;
+  FusionEngineFramer framer_ = FusionEngineFramer(FRAMER_BUFFER_BYTES);
+  LbandCallback callback_ = nullptr;
+};
+
+/******************************************************************************/
+int main(int argc, const char* argv[]) {
+  if (argc != 2) {
+    printf("Usage: %s FILE\n", argv[0]);
+    printf(R"EOF(
```
adamshapiro0, line 61: R string seems a little unnecessary here?

examples/lband_decode/lband_decode.cc
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * @brief Specify a function to be called when L-band data is framed.
+   *
+   * @param callback The function to be called with the message payload.
+   */
+  void SetMessageCallback(LbandCallback callback) { callback_ = callback; }
+
+ private:
+  static constexpr size_t FRAMER_BUFFER_BYTES = 1024;
+  FusionEngineFramer framer_ = FusionEngineFramer(FRAMER_BUFFER_BYTES);
```
adamshapiro0, line 53: Doing a default assign of a non-primitive in the class definition is pretty uncommon, initializer list might be slightly more clear to some readers.

examples/lband_decode/lband_decode.cc
```diff
@@ -0,0 +1,99 @@
...
+#include <fstream>
+
+#include <point_one/fusion_engine/messages/gnss_corrections.h>
+#include <point_one/fusion_engine/parsers/fusion_engine_framer.h>
+
+using namespace point_one::fusion_engine::messages;
+using namespace point_one::fusion_engine::parsers;
+
+/******************************************************************************/
+class LbandDecoder {
```
adamshapiro0, line 17: I'd recommend removing this class entirely for clarity. It actually complicates the understanding a little bit here since now the app is creating a class and registering two different callbacks, and the reader needs to follow why. Plus, `ProcessFEInput()` registers a new callback with the framer on every read call, which is not really recommended.

Instead, I'd suggest simply defining the framer in `main()`, and then creating a callback directly to the framer that does the combined "is this an L-band message?" and "print and write to disk". Then the while loop would just call `OnData()`. `main()` would look almost exactly the same as it already does, and this class wouldn't exist at all, so the app would be much simpler to follow.

adamshapiro0 (changes requested):
Looks like it should work, but I think the decoder class complicates the understanding a fair amount without really adding anything. I recommend simplifying the app.

examples/lband_decode/lband_decode.cc
axlan: The idea was that since this is made as an example for customer they could copy and paste the class without needing to dig into the internals. It basically just hides all the FE logic, and would make more sense as it's own file.

axlan:
@adamshapiro0 Any idea about the MSVC error I'm getting:
```
 [100%] Linking CXX executable lband_decode.exe
LINK Pass 1: command "C:\PROGRA~1\MICROS~2\2022\ENTERP~1\VC\Tools\MSVC\1435~1.322\bin\HostX64\x64\link.exe /nologo @CMakeFiles\lband_decode.dir\objects1.rsp /out:lband_decode.exe /implib:lband_decode.lib /pdb:D:\a\fusion-engine-client\fusion-engine-client\cmake_build\examples\lband_decode\lband_decode.pdb /version:0.0 /machine:x64 /debug /INCREMENTAL /subsystem:console ..\..\fusion_engine_client.lib kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib /MANIFEST /MANIFESTFILE:CMakeFiles\lband_decode.dir/intermediate.manifest CMakeFiles\lband_decode.dir/manifest.res" failed (exit code 1120) with the following output:
lband_decode.cc.obj : error LNK2019: unresolved external symbol "public: __cdecl point_one::fusion_engine::parsers::FusionEngineFramer::FusionEngineFramer(void *,unsigned __int64)" (??0FusionEngineFramer@parsers@fusion_engine@point_one@@QEAA@PEAX_K@Z) referenced in function "public: __cdecl point_one::fusion_engine::parsers::FusionEngineFramer::FusionEngineFramer(unsigned __int64)" (??0FusionEngineFramer@parsers@fusion_engine@point_one@@QEAA@_K@Z)
lband_decode.cc.obj : error LNK2019: unresolved external symbol "public: unsigned __int64 __cdecl point_one::fusion_engine::parsers::FusionEngineFramer::OnData(unsigned char const *,unsigned __int64)" (?OnData@FusionEngineFramer@parsers@fusion_engine@point_one@@QEAA_KPEBE_K@Z) referenced in function "public: void __cdecl LbandDecoder::ProcessFEInput(unsigned char const *,unsigned __int64)" (?ProcessFEInput@LbandDecoder@@QEAAXPEBE_K@Z)
lband_decode.exe : fatal error LNK1120: 2 unresolved externals
NMAKE : fatal error U1077: 'C:\hostedtoolcache\windows\cmake\3.16.9\x64\cmake-3.16.9-win32-x86\bin\cmake.exe -E vs_link_exe --intdir=CMakeFiles\lband_decode.dir --rc=C:\PROGRA~2\WI3CF2~1\10\bin\100226~1.0\x64\rc.exe --mt=C:\PROGRA~2\WI3CF2~1\10\bin\100226~1.0\x64\mt.exe --manifests  -- C:\PROGRA~1\MICROS~2\2022\ENTERP~1\VC\Tools\MSVC\1435~1.322\bin\HostX64\x64\link.exe /nologo @CMakeFiles\lband_decode.dir\objects1.rsp @C:\Users\RUNNER~1\AppData\Local\Temp\nmDD63.tmp' : return code '0xffffffff'
Stop.
NMAKE : fatal error U1077: '"C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.35.32215\bin\HostX64\x64\nmake.exe" -s -f examples\lband_decode\CMakeFiles\lband_decode.dir\build.make /nologo -SL                 examples\lband_decode\CMakeFiles\lband_decode.dir\build' : return code '0x2'
Stop.
NMAKE : fatal error U1077: '"C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.35.32215\bin\HostX64\x64\nmake.exe" -s -f CMakeFiles\Makefile2 /nologo -S                  all' : return code '0x2'
Stop.
Error: Process completed with exit code 2.
```
My guess is that since the parser isn't referenced by other parts of the library it's being stripped out of the DLL. Is there a reason that we're using `PRIVATE` for:
```
if (MSVC)
    target_compile_definitions(fusion_engine_client PRIVATE BUILDING_DLL)
endif()
```

examples/lband_decode/lband_decode.cc
adamshapiro0: Yeah that's what I figured. Makes sense, but in practice I think it actually just makes things more confusing with the multiple callbacks and stuff based on my experience with how people have used these examples in the past.

adamshapiro0:
Hmm not sure. I don't honestly remember what the deal was with that MSVC case it's been so long since I got it building back then.

axlan:
Ugh, it's actually pretty complicated. The issue is that the parser is missing `P1_EXPORT`, but adding it leads to this issue: https://stackoverflow.com/questions/2132747/warning-c4251-when-building-a-dll-that-exports-a-class-containing-an-atlcstrin

examples/lband_decode/lband_decode.cc
```diff
@@ -0,0 +1,68 @@
...
+
+/******************************************************************************/
+int main(int argc, const char* argv[]) {
+  if (argc != 2) {
+    printf("Usage: %s FILE\n", argv[0]);
+    printf("Decode L-band corrections and write contents to 'lband.bin'.");
+    return 0;
+  }
+
+  char buffer[2048];
```
adamshapiro0, line 39: Minor suggestion: declare this down above the while loop where it's used so there's no confusion between this buffer size and the framing buffer size, which are unrelated.

examples/lband_decode/lband_decode.cc
```diff
@@ -0,0 +1,68 @@
...
+  static constexpr size_t FRAME_BUFFER_BYTES = 1024;
+  FusionEngineFramer framer(FRAME_BUFFER_BYTES);
+
+  std::ifstream in_stream(argv[1], std::ifstream::binary);
+  if (!in_stream.is_open()) {
+    printf("Error opening file '%s'.\n", argv[1]);
+    return 1;
+  }
+  out_stream.open("lband.bin", std::ifstream::binary);
```
adamshapiro0, line 48: Might be nice to make the output file/path an optional second user arg?

adamshapiro0 (approved):
Looks good, couple minor suggestions left

PR #246, opened 2023-08-15
Add P1 time preceding reset to event log

wbrannon:
# New Features
- Adds column to event log that displays the last valid P1 time (if available) before all resets

### DIFF
--- a/python/fusion_engine_client/analysis/analyzer.py
+++ b/python/fusion_engine_client/analysis/analyzer.py
@@ -1913,7 +1913,7 @@ Gold=Float, Green=Integer (Not Fixed), Blue=Integer (Fixed, Float Solution Type)
             rows.append([
                 f'{(system_time_ns - system_t0_ns) / 1e9:.3f}' if system_time_ns is not None else 'N/A',
                 f'{system_time_ns / 1e9:.3f}' if system_time_ns is not None else 'N/A',
-                'N/A',
+                '',
                 event_type.to_string(include_value=True),
                 f'0x{flags:016X}' if flags is not None else 'N/A',
                 description_str.replace('<', '[').replace('>', ']').replace('\n', '<br>'),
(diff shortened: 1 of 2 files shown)


wbrannon:
After these changes:

**Event log before:**

image

**Event log after:**

image


Now, with more resets:
**Before:** 

image

**After:**

image

python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -479,6 +479,29 @@ def _print_progress(self, file_size=None):
...
+        """
+        # Jump to offset governed by index.
+        self.input_file.seek(index.offset, os.SEEK_SET)
+
+        # Generate header and payload.
+        data = self.input_file.read(MessageHeader.calcsize())
+        header = MessageHeader()
+        header.unpack(data, warn_on_unrecognized=False)
+        payload_bytes = self.input_file.read(header.payload_size_bytes)
+        cls = message_type_to_class.get(header.message_type, None)
```
adamshapiro0, line 499: This can return `None` if the class is not recognized. If that happens, you should return the `payload_bytes` since you can't call `cls()`.

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -1920,6 +1931,30 @@ def plot_events(self):
...
+        times_before_resets = {}
+        file_index = self.reader.get_index()
+        for entry in file_index[::-1]:
+            if entry.type == MessageType.EVENT_NOTIFICATION or get_time_before_reset:
+                # Parse entry at index for payload.
+                header, payload = self.reader.reader.parse_entry_at_index(entry)
+                system_time_ns = payload.get_system_time_ns()
+
+                if get_time_before_reset and payload.get_p1_time() is not None:
+                    times_before_resets[curr_reset_time] = payload.get_p1_time().seconds
```
adamshapiro0, line 1948: Better to use `float(payload.get_p1_time())` to get at the second value. That way, if we change the internal definition in the future (e.g., store integer and fractional parts separately), this code doesn't need to know about that.

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -1907,11 +1913,16 @@ def plot_events(self):
             rows.append([
                 f'{(system_time_ns - system_t0_ns) / 1e9:.3f}' if system_time_ns is not None else 'N/A',
                 f'{system_time_ns / 1e9:.3f}' if system_time_ns is not None else 'N/A',
+                'N/A',
```
adamshapiro0, line 1916: I'd actually just leave this col empty string by default since like 99% of the stuff in the table will not have this column. Makes the reset events stand out more.

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -1920,6 +1931,30 @@ def plot_events(self):
...
+        curr_reset_time = None
+        get_time_before_reset = False
+
+        times_before_resets = {}
+        file_index = self.reader.get_index()
+        for entry in file_index[::-1]:
+            if entry.type == MessageType.EVENT_NOTIFICATION or get_time_before_reset:
+                # Parse entry at index for payload.
+                header, payload = self.reader.reader.parse_entry_at_index(entry)
+                system_time_ns = payload.get_system_time_ns()
```
adamshapiro0, line 1945: Move below - you only use this if you hit a reset event in the if below.

adamshapiro0 (approved):
Looks good. Make sure to test it on a log with multiple resets.

PR #250, opened 2023-08-16
Add back dropped portability fix for Keil ARMCC NAN macro.

axlan:


adamshapiro0 (commented):
What is this macro used for? I removed this the other day when I was fixing the other ARMCC typedef issues because it doesn't appear to be called anywhere in this repo.

axlan:
As mentioned in the link in the comment, __ESCAPE__ is used in the keil C headers as part of declaring the `NAN` macro. Their preprocessor doesn't handle it correctly and it results in compiler errors when the `NAN` macro is used. `__ESCAPE__` is meant to be used when using the kiel include files in other compilers so removing the macro functionality has no affect besides fixing the `NAN` macro bug.

adamshapiro0:
Oh ok I see. I missed that. I'd suggest actually explaining that in the comment in our code, particularly that this macro is not used in our repo but is used in Keil's `math.h` header where they define `NAN`.

adamshapiro0 (approved):

PR #252, opened 2023-08-29
Fixed slow read times for logs with no P1 or system time entries.

adamshapiro0:
# Changes
- Only check for t0 once per log
- Allow `--plot` argument to be specified multiple times


wbrannon (approved):
Looks good.

PR #253, opened 2023-08-30
Added reset recovery timing plot.

adamshapiro0:
# New Features
- Added plot of reset recovery timing (time until reacquisition and navigation)
- Added `return_message_index` option to report the location of each message within the file
- Added `MixedLogReader` option to seek to a specific message by index

# Changes
- Allow multiple instances of `p1_display --plot NAME` argument
- Use `--time-axis=absolute` to display absolute P1/system time by default instead of relative time

### DIFF
--- a/python/fusion_engine_client/analysis/analyzer.py
+++ b/python/fusion_engine_client/analysis/analyzer.py
@@ -391,9 +391,7 @@ class Analyzer(object):
                 except StopIteration:
                     break
 
-                if next_reset_index is not None and pose_index >= next_reset_index:
-                    break
-                elif invalid_p1_time is None:
+                if invalid_p1_time is None:
                     if message.solution_type == SolutionType.Invalid:
                         invalid_p1_time = message.get_p1_time()
                         invalid_p1_time_sec = float(invalid_p1_time)
--- a/python/fusion_engine_client/analysis/analyzer.py
+++ b/python/fusion_engine_client/analysis/analyzer.py
@@ -340,6 +340,111 @@ class Analyzer(object):
 
         self._add_figure(name="time_scale", figure=figure, title="Time Scale")
 
+    def plot_reset_timing(self):
+        if self.output_dir is None:
+            return
+
+        # Find reset events.
+        result = self.reader.read(message_types=[EventNotificationMessage], return_message_index=True, **self.params)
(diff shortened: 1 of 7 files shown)


python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -340,6 +340,126 @@ def _calc_stats(input):
...
+            # Find the pose where the solution went invalid after the reset, then where it went valid after that.
+            invalid_p1_time = None
+            valid_p1_time = None
+            while True:
+                try:
+                    _, message, pose_index = self.reader.read_next(return_message_index=True)
+                except StopIteration:
+                    break
+
+                if next_reset_index is not None and pose_index >= next_reset_index:
```
axlan, line 394: Is this condition possible since you've already filtered the log to not include times past `next_reset_index`?

python/fusion_engine_client/parsers/file_index.py
```diff
@@ -121,7 +121,7 @@ class FileIndex(object):
     #   timestamp is its exact time
     _RAW_DTYPE = np.dtype([('int', '<u4'), ('type', '<u2'), ('offset', '<u8')])
 
-    _DTYPE = np.dtype([('time', '<f8'), ('type', '<u2'), ('offset', '<u8')])
+    _DTYPE = np.dtype([('time', '<f8'), ('type', '<u2'), ('offset', '<u8'), ('message_index', '<u8')])
```
axlan, line 124: I'm a little confused why this needs to be written out into the binary. Can't the `message_index` be determined by the offset you're at into the index? So if each entry is 11 bytes, you know the index is just index_file.tell()/11.

Or if it's being read as a numpy array, it's just the array's index, unless I'm missing something?

axlan: It makes sense to preserve the index in the reader when you filter (pandas does something similar, but it seems odd to write it to the index file)

axlan (approved):
LGTM

python/fusion_engine_client/analysis/analyzer.py
adamshapiro0: Nope don't think so. I added this before I added the slice filtering above. Will remove.

PR #254, opened 2023-08-30
Plot min/max when decimating time interval plot.

adamshapiro0:
# Changes
- Plot min/max when decimating time interval plot to avoid missing outliers when looking at a long log


wbrannon (approved):
Looks good.

PR #258, opened 2023-09-18
Delete `WHEEL_SPEED` `WheelSensorType` enum

wbrannon:


anathan:
Does this effect quectel configs at all?

wbrannon:
@anathan I think the `WHEEL_SPEED` enum will affect Quectel configs the exact same way that it affects Atlas configs: if the `wheel_sensor_type` config parameter is set to `WHEEL_SPEED`, we _won't_ get wheel speeds as we might expect.

PR #259, opened 2023-09-21
Fixed crash while plotting corrected vehicle/wheel speeds.

adamshapiro0:
# Changes
- Include measurement time/source in corrected measurement numpy output for API consistency with raw measurements
- Renamed DOP figure to group with other GNSS figures


axlan (approved):
LGTM

PR #261, opened 2023-09-22
Removed p1_print index disabling, and updated summary table.

adamshapiro0:
# Changes
- Don't disable/delete the log index file in `p1_print` when reading messages with system time (input measurements, etc.)
- Include the message type number in the `p1_print -s` summary table


axlan (approved):
LGTM

PR #262, opened 2023-09-25
Delete `TICK_RATE` enum from `WheelSensorType`

wbrannon:


src/point_one/fusion_engine/messages/configuration.h
```diff
@@ -856,12 +856,7 @@ struct P1_ALIGNAS(4) VehicleDetails {
 enum class WheelSensorType : uint8_t {
   /** Wheel/vehicle speed data not available. */
   NONE = 0,
-  /**
-   * Individual wheel rotation rates, reported as an encoder tick rate (in
-   * ticks/second). Will be scaled to meters/second using the specified scale
-   * factor.
-   */
-  TICK_RATE = 1,
+  RESERVED = 1,
```
adamshapiro0, line 859: I think we should omit these from the actual enum. We can leave them in here (and in Python) just to act as a visual separator and for information purposes, but let's go ahead and comment them out so they're not actually usable.

We will want to make sure they're marked as reserved in the protocol spec, but that's separate.

PR #263, opened 2023-09-26
Updated cmake install action.

adamshapiro0:
As of 9/26/2023, the old version can no longer install CMake 3.16.x and gives the following error:

```
Run jwlawson/actions-setup-cmake@v1.4
  with:
    cmake-version: 3.16.x
    github-api-token: ***
Error: Unable to find version matching 3.16.x
```


adamshapiro0:
@axlan FYI

axlan (approved):
LGTM

PR #265, opened 2023-09-28
Make FileIndex flexible for different type keys.

axlan:


python/fusion_engine_client/parsers/file_index.py
```diff
@@ -191,7 +195,7 @@ def load(self, index_path, data_path=None, delete_on_error=True):
             if not os.path.exists(data_path):
                 # If the user didn't explicitly set data_path and the default file doesn't exist, it is not considered
                 # an error.
-                if self._data['type'][-1] == MessageType.INVALID:
+                if self._data['type'][-1] == self.enum_class.INVALID:
```
adamshapiro0, line 198: `INVALID` is just the name used by `MessageType`. Another enum might define a different value to use as an undefined/invalid value. Probably should pass this into the constructor too.

adamshapiro0 (approved):
Looks good, just careful with `INVALID`.

PR #266, opened 2023-09-29
Add more fields to Python IMU example, and add wheel speed one.

axlan:
# Changes
 - Add `DataLoader.convert_to_gps_time`
 - Add GPS Time and raw data to `extract_imu_data.py`
 - Add GPS Time and raw data to `extract_wheel_speed_data.py`

### DIFF
--- a/python/examples/extract_wheel_speed_data.py
+++ b/python/examples/extract_wheel_speed_data.py
@@ -17,7 +17,7 @@ from fusion_engine_client.utils.argument_parser import ArgumentParser
 if __name__ == "__main__":
     # Parse arguments.
     parser = ArgumentParser(description="""\
-Extract IMU accelerometer and gyroscope measurements.
+Extract wheel speed data.
 """)
     parser.add_argument('--log-base-dir', metavar='DIR', default=DEFAULT_LOG_BASE_DIR,
                         help="The base directory containing FusionEngine logs to be searched if a log pattern is "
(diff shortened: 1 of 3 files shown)


python/fusion_engine_client/analysis/data_loader.py
```diff
@@ -649,14 +649,15 @@ def get_log_reader(self) -> MixedLogReader:
     def get_input_path(self):
         return self.reader.input_file.name
 
-    def convert_to_p1_time(self,
-                           times: Union[Iterable[Union[datetime, gpstime, Timestamp, float]],
-                                        Union[datetime, gpstime, Timestamp, float]],
-                           assume_utc: bool = False) ->\
+    def _convert_time(self, gps_to_p1,
```
adamshapiro0, line 652: I'd suggest we make this a pair of enum or a string args `source_type` and `target_type` instead: `source_type='p1', target_type='gps'`, etc. Having `gps_to_p1=False` mean that it will convert p1->gps instead is a little unintuitive, and in the future we plan to also add conversion between system time and P1/GPS time.

python/fusion_engine_client/analysis/data_loader.py
```diff
@@ -739,24 +741,84 @@ def _to_gps_or_p1(value):
...
+                    time_sec[gps_idx] = np.nan
+                else:
+                    # The relationship between P1 time and GPS time should not change rapidly since P1 time is rate-steered
+                    # to align with GPS time. As a result, it should be safe to extrapolate between gaps in P1 or GPS times,
+                    # as long as the gaps aren't extremely large. NumPy's interp() function does not extrapolate, so we
+                    # instead use SciPy's function.
+                    f = sp.interpolate.interp1d(gps_ref_sec, p1_ref_sec, fill_value='extrapolate')
+                    time_sec[gps_idx] = f(time_sec[gps_idx])
+        else:
+            p1_idx = not is_gps_time(time_sec)
```
adamshapiro0, line 758: `is_gps_time()` returns a boolean numpy array. `not` won't work on a logical array, you need to use `~`.

python/fusion_engine_client/analysis/data_loader.py
```diff
@@ -739,24 +741,84 @@ def _to_gps_or_p1(value):
...
+                 - Note that UTC timestamps cannot be specified `float` unless `assume_utc == True`
+        @param assume_utc If `True`:
+               - For `float` values, assume values greater than the POSIX offset for 2000/1/1 are UTC timestamps in
+                 seconds.
+               - For `datetime`, if `tzinfo` is not set, assume it is `timezone.utc`. Otherwise, interpret the timestamp
+                 in the local timezone.
+
+        @return A numpy array containing P1 time values (in seconds), or `nan` if the value could not be converted.
+        """
+        return self._convert_time(True, times, assume_utc)
```
adamshapiro0, line 795: Use named args here and below

adamshapiro0 (approved):
Looks good, but I think we should change the source/target type interface so we have more flexibility in the future.

python/examples/extract_wheel_speed_data.py
```diff
@@ -0,0 +1,99 @@
...
+        sys.exit(1)
+
+    if log_id is None:
+        logger.info('Loading %s.' % os.path.basename(input_path))
+    else:
+        logger.info('Loading %s from log %s.' % (os.path.basename(input_path), log_id))
+
+    # Read satellite data from the file.
+    reader = DataLoader(input_path)
+    result = reader.read(message_types=[WheelSpeedOutput, RawWheelSpeedOutput], show_progress=True)
```
adamshapiro0, line 52: This script should also support `[Raw]VehicleSpeedOutput` since that's a very common use case for hardware wheel ticks.

python/fusion_engine_client/analysis/data_loader.py
axlan: Ah, this actually isn't quite right. It can return a bool or a numpy array. In my script it returns a bool so `~` actually won't work.

PR #267, opened 2023-10-04
[FUS-1585] Added fast indexing support in Python.

axlan:
# Changes
 - Speed up indexing to ~100MB/s on a 16 core laptop with SSD
 - Make the MixedLogReader assume it will index the file on construction

# Bug Fixes
 - Fix p1_print for unknown types

This generally will make indexing faster then iterating through a log with MixedLogReader  with the index file.

### DIFF
--- a/python/fusion_engine_client/analysis/data_loader.py
+++ b/python/fusion_engine_client/analysis/data_loader.py
@@ -135,15 +135,15 @@ class DataLoader(object):
 
     logger = logging.getLogger('point_one.fusion_engine.analysis.data_loader')
 
-    def __init__(self, path=None, generate_index=True, ignore_index=False):
+    def __init__(self, path=None, save_index=True, ignore_index=False):
         """!
         @brief Create a new reader instance.
 
         @param path The path to a binary file containing FusionEngine messages, or an existing Python file object.
-        @param generate_index If `True`, generate a `.p1i` index file if one does not exist for faster reading in the
+        @param save_index If `True`, save a `.p1i` index file if one does not exist for faster reading in the
                future. See @ref FileIndex for details.
         @param ignore_index If `True`, ignore the existing index file and read from the `.p1log` binary file directly.
-               If `generate_index == True`, this will delete the existing file and create a new one.
+               If `save_index == True`, this will delete the existing file and create a new one.
         """
         self.reader: MixedLogReader = None
 
@@ -155,26 +155,24 @@ class DataLoader(object):
         self._need_t0 = True
         self._need_system_t0 = True
 
-        self._generate_index = generate_index
+        self._generate_index = save_index
         if path is not None:
-            self.open(path, generate_index=generate_index, ignore_index=ignore_index)
+            self.open(path, save_index=save_index, ignore_index=ignore_index)
 
-    def open(self, path, generate_index=True, ignore_index=False):
+    def open(self, path, save_index=True, ignore_index=False):
         """!
         @brief Open a FusionEngine binary file.
 
         @param path The path to a binary file containing FusionEngine messages, or an existing Python file object.
-        @param generate_index If `True`, generate a `.p1i` index file if one does not exist for faster reading in the
+        @param save_index If `True`, generate a `.p1i` index file if one does not exist for faster reading in the
                future. See @ref FileIndex for details.
         @param ignore_index If `True`, ignore the existing index file and read from the `.p1log` binary file directly.
-               If `generate_index == True`, this will delete the existing file and create a new one.
+               If `save_index == True`, this will delete the existing file and create a new one.
         """
         self.close()
 
-        self.reader = MixedLogReader(input_file=path, generate_index=generate_index, ignore_index=ignore_index,
+        self.reader = MixedLogReader(input_file=path, save_index=save_index, ignore_index=ignore_index,
                                      return_bytes=True, return_message_index=True)
-        if self.reader.have_index():
-            self.logger.debug("Using index file '%s'." % self.reader.index_path)
 
         # Read the first message (with P1 time) in the file to set self.t0.
         #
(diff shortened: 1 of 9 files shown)


python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -102,38 +102,13 @@ def __init__(self, input_file, warn_on_gaps: bool = False, show_progress: bool =
...
-                                                  len(np.unique(self.index.type))
-                except ValueError as e:
-                    self.logger.error("Error loading index file: %s" % str(e))
-            else:
-                self.logger.debug("No index file found @ '%s'." % self.index_path)
 
-        self.index_builder = None
-        self.set_generate_index(generate_index)
+        # Open the companion index file if one exists, otherwise index the file.
+        self._original_index = fast_indexer.fast_generate_index(input_path, ignore_index, generate_index, max_bytes)
```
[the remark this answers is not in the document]
adamshapiro0: (And update the function doxy to clarify behavior of renamed `save_index` and `ignore_index` arguments)

python/tests/test_mixed_log_reader.py
```diff
@@ -220,10 +217,8 @@ def test_read_no_generate_index(self, data_path):
         expected_messages = [m for m in messages if isinstance(m, PoseMessage)]
 
         reader = MixedLogReader(str(data_path), generate_index=False)
```
adamshapiro0, line 219: Recommend renaming this arg to `save_index`, but also this unit test doesn't really make sense anymore as written. If we want to change it to a test for the `save_index` arg, that's fine, but then we need to check that the index file does not exist after this call, and then that the check results call below still works correctly.

python/tests/test_mixed_log_reader.py
```diff
@@ -256,16 +251,11 @@ def test_read_overwrite_index(self, data_path):
...
-        # deleted the existing one.
-        assert reader.index is None
-        assert not os.path.exists(reader.index_path)
-
         reader.filter_in_place((PoseMessage,))
         self._check_results(reader, expected_messages)
 
         # Check that we generated a new index.
-        assert os.path.exists(reader.index_path)
+        assert os.path.exists(FileIndex.get_path(data_path))
```
adamshapiro0, line 258: Does `index_path` not exist anymore? Any particular reason?

python/fusion_engine_client/analysis/data_loader.py
```diff
@@ -173,8 +173,6 @@ def open(self, path, generate_index=True, ignore_index=False):
 
         self.reader = MixedLogReader(input_file=path, generate_index=generate_index, ignore_index=ignore_index,
```
adamshapiro0, line 174: Recommend renaming `generate_index` -> `save_index` in this class too

python/fusion_engine_client/parsers/fast_indexer.py
```diff
@@ -0,0 +1,185 @@
...
+
+# NOTE: This needs to be larger then the biggest possible FE message. The
+# smaller it is, the faster the indexer can run, but if it's not large enough,
+# messages may be missed.
+_MAX_FE_MSG_SIZE_BYTES = 1024 * 12
+
+# This has been tuned on my laptop to where increasing the read size was giving
+# diminishing returns on speed.
+_READ_SIZE_BYTES = 80 * 1024
+_READ_WORDS = int(_READ_SIZE_BYTES / 2)
```
adamshapiro0, line 21: To be clear, by word in this context I think you mean "2 byte chunk that might have a preamble" and not word in the normal sense of like 32b value, correct? Might suggest renaming to avoid confusion.

python/fusion_engine_client/parsers/fast_indexer.py
```diff
@@ -0,0 +1,185 @@
...
+
+    This basic logic would be relatively easy to extend to do general parallel processing on messages.
+
+    @param input_path The path to the file to be read.
+    @param force_reindex If `True` regenerate the index even if there's an existing file.
+    @param save_index If `True` save the index to disk after generation.
+    @param max_bytes If specified, read up to the maximum number of bytes.
+    @param block_starts The blocks of data to search for FE messages in.
+
+    @return The loaded or generated @ref FileIndex
```
adamshapiro0, line 126: Missing .

python/fusion_engine_client/parsers/fast_indexer.py
```diff
@@ -0,0 +1,185 @@
...
+
+    _logger.info(f'Indexing file "{input_path}". This may take a few seconds.')
+    _logger.debug(f'Using Threads: {num_threads}')
+
+    # These are the args passed to the _search_blocks_for_fe instances.
+    args: list[Tuple[str, List[int]]] = []
+    # Allocate which blocks of bytes will be processed by each thread.
+    num_blocks = math.ceil(file_size / _READ_SIZE_BYTES)
+    # Each thread will process at least blocks_per_thread blocks. If the number
+    # doesn't device evenly, distribute the remainder.
```
adamshapiro0, line 156: divide

python/fusion_engine_client/parsers/fast_indexer.py
```diff
@@ -0,0 +1,185 @@
...
+        # Kick off the threads to process with their args. Then concatenate their returned data.
+        index_raw = np.concatenate([o for o in p.starmap(_search_blocks_for_fe, args)])
+
+    total_entries = len(index_raw)
+
+    _logger.debug(f'FE messages found: {total_entries}')
+
+    # This does an unnecessary conversion back and forth from the _RAW_DTYPE to
+    # the _DTYPE and back again when saving. This is to avoid needing to deal
+    # with the EOF entry directly. This adds less then a second, and could be
```
adamshapiro0, line 179: than

adamshapiro0 (approved):
Looks good

python/tests/test_mixed_log_reader.py
axlan: It generally doesn't make sense to reference the file directly. It also might be confusing in the case where you specified not to save it, but you did load it.

python/fusion_engine_client/parsers/fast_indexer.py
```diff
@@ -0,0 +1,185 @@
...
+                    p1_time = Timestamp()
+                    # About half the indexing time is spent doing this p1_time check. This would
+                    # probably take some doing to optimize. One approach to speed up the validation
+                    # and p1 time checks is to read the message as a uint32_t and having a map to
+                    # the p1_time (if present) for each class.
+                    cls = message_type_to_class.get(header.message_type, None)
+                    if cls is not None:
+                        try:
+                            payload = cls()
+                            if hasattr(payload, 'p1_time') or hasattr(payload, 'details'):
```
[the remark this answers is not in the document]
axlan: I mention that in the comment right above here. I don't think it's worth the effort for maintaining that for now.

PR #270, opened 2023-10-05
[FUS-1598] Add C++ request_version example.

axlan:


examples/request_version/request_version.cc
```diff
@@ -0,0 +1,116 @@
...
+
+  header->crc = CalculateCRC(storage);
+
+  printf("Sending VersionInfoMessage request:\n  ");
+  // This data would be sent over serial to the device.
+  PrintHex(storage, sizeof(MessageHeader) + sizeof(MessageRequest));
+  printf("\n");
+
+  //////////////////////////////////////////////////////////////////////////////
+  // Generate an example response
```
anathan, line 60: I know you make the comment later, but its probably worth explaining up here that you're generating this example response that would normally be generated by the hardware for clarity.

examples/request_version/request_version.cc
```diff
@@ -0,0 +1,116 @@
...
+  auto req_message = reinterpret_cast<MessageRequest*>(buffer);
+  *req_message = MessageRequest();
+
+  req_message->message_type = VersionInfoMessage::MESSAGE_TYPE;
+
+  header->crc = CalculateCRC(storage);
+
+  printf("Sending VersionInfoMessage request:\n  ");
+  // This data would be sent over serial to the device.
+  PrintHex(storage, sizeof(MessageHeader) + sizeof(MessageRequest));
```
anathan, line 56: Perhaps we make a stub function called "SendData(void* data, size_t len)" that doesn't do anything but is the obvious plug in for where you're serial port/ethernet port/etc. would go?

anathan (approved):

examples/request_version/request_version.cc
```diff
@@ -0,0 +1,137 @@
+/**************************************************************************/ /**
+* @brief Message encode example.
```
adamshapiro0, line 2: Update this to explain what this example app does, and also add the app to https://github.com/PointOneNav/fusion-engine-client/tree/master#example-applications

PR #275, opened 2023-10-12
Install all applications as console scripts via setup.py.

adamshapiro0:
This should install them correctly in Windows and Mac, not just Linux. That way users can run:
```
> p1_display path/to/log
```
at the command line in all operating systems.


python/README.md
```diff
@@ -64,6 +57,14 @@ FusionEngine message specification.
           - This class is used by the `bin/p1_display` application
         - [file_reader.py](analysis/file_reader.py) - `FileReader` class, capable of loading and time-aligning
           FusionEngine data captured in a `*.p1log` file
+      - `applications` - Applications and processing tools
+        - [p1_display](bin/p1_display) - Generate HTML plots of vehicle trajectory, etc. (runs `analyzer.py`)
+        - [p1_extract](bin/p1_extract) - Extract FusionEngine message contents from a binary file containing mixed data
+          (e.g., interleaved RTCM and FusionEngine messages)
+        - [p1_lband_extract](bin/p1_lband_extract) - Extract L-band data bits contained in a log of FusionEngine
```
anathan, line 64: isnt this in applications now?

python/README.md
```diff
@@ -64,6 +57,14 @@ FusionEngine message specification.
...
+          `LBandFrameMessage`
+        - [p1_print](bin/p1_print) - Print the contents of FusionEngine messages found in a binary file to the
```
anathan, line 66: same.

python/README.md
```diff
@@ -97,12 +97,13 @@ FusionEngine message specification.
    ```
 3. Install the `fusione-engine-client` module, including all analysis and data processing tools:
    ```bash
-   python3 -m pip install -e /path/to/fusion-engine-client[all]
+   python3 -m pip install -e /path/to/fusion-engine-client
```
anathan, line 100: yes!

anathan (approved):
just check readme comment, otherwise g2g.

python/README.md
adamshapiro0: Good catch, fixed and reorganized this stuff a little better

PR #276, opened 2023-10-12
Add timestamp and height parameters to `extract_position_data.py` KML output

wbrannon:
# Changes
- Associates colors defined in `analyzer.py` to KML coordinates
- Creates placemark for each coordinate
- Adds timestamp and MSL height parameters to coordinates

### DIFF
--- a/python/examples/extract_position_data.py
+++ b/python/examples/extract_position_data.py
@@ -81,8 +81,8 @@ KML_TEMPLATE_END = """\
 
 KML_TEMPLATE = """\
     <Placemark>
-      <Timestamp>%(timestamp)s</Timestamp>
-      <styleUrl>#type-%(solution_type)s</styleUrl>
+      <TimeStamp><when>%(timestamp)s</when></TimeStamp>
+      <styleUrl>#type-%(solution_type)d</styleUrl>
       <Point>
         <altitudeMode>absolute</altitudeMode>
         <coordinates>%(coordinates)s</coordinates>
@@ -158,9 +158,14 @@ Extract position data to both CSV and KML files.
     with open(path, 'w') as f:
         f.write(KML_TEMPLATE_START)
         for pose in pose_data.messages:
+          # IMPORTANT: KML heights are specified in MSL, so we convert the ellipsoid heights to orthometric below using
+          # the reported geoid undulation (geoid height). Undulation values come from a geoid model, and are not
+          # typically precise. When analyzing position performance compared with another device, we strongly recommend
+          # that you do the performance using ellipsoid heights. When comparing in MSL, if the geoid models used by the
+          # two devices are not exactly the same, the heights may differ by multiple meters.
           f.write(KML_TEMPLATE %
-                  {'timestamp': '\n'.join([str(pose.gps_time.as_utc())]),
-                  'solution_type':'\n'.join([str(int(pose.solution_type))]),
+                  {'timestamp': '\n'.join([str(pose.gps_time.as_utc().isoformat())]),
+                  'solution_type': int(pose.solution_type),
                   'coordinates': '\n'.join(['%.8f,%.8f,%.8f' % (pose.lla_deg[1], pose.lla_deg[0], pose.lla_deg[2] - pose.undulation_m)])})
         f.write(KML_TEMPLATE_END)
 
--- a/python/examples/extract_position_data.py
+++ b/python/examples/extract_position_data.py
@@ -91,6 +91,20 @@ KML_TEMPLATE = """\
     </Placemark>
 """
 
+KML_TEMPLATE_LOOKAT = """\
+    <LookAt>
+      <latitude>%(latitude).8f</latitude>
+      <longitude>%(longitude).8f</longitude>
+      <altitude>%(altitude).8f</altitude>
+      <altitudeMode>absolute</altitudeMode>
+      <range>250</range>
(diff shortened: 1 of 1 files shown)


python/examples/extract_position_data.py
```diff
@@ -14,30 +14,82 @@
...
+
+KML_TEMPLATE_END = """\
   </Document>
 </kml>
 """
 
+KML_TEMPLATE = """\
+    <Placemark>
+      <Timestamp>%(timestamp)s</Timestamp>
+      <styleUrl>#type-%(solution_type)s</styleUrl>
```
adamshapiro0, line 85: use d instead of s, then you don't have to convert to `str()` below, you can just use the integer directly.

python/examples/extract_position_data.py
```diff
@@ -104,9 +156,12 @@
...
     logger.info("Generating '%s'." % path)
     with open(path, 'w') as f:
-        f.write(KML_TEMPLATE %
-                {'coordinates': '\n'.join(['%.8f,%.8f' % (message.lla_deg[1], message.lla_deg[0])
-                                           for message in pose_data.messages
-                                           if message.solution_type != SolutionType.Invalid])})
+        f.write(KML_TEMPLATE_START)
+        for pose in pose_data.messages:
+          f.write(KML_TEMPLATE %
+                  {'timestamp': '\n'.join([str(pose.gps_time.as_utc())]),
```
adamshapiro0, line 162: This is not the correct timestamp format for KML. `as_utc()` returns a `datetime` object, and `str()` of that returns a string like `2023-10-05 21:22:54+00:00`, but looking at https://developers.google.com/kml/documentation/time#how-to-specify-time and https://developers.google.com/kml/documentation/time#example:-showing-movement-of-a-point-along-a-path, they're expecting a timestamp like `2007-01-14T21:05:02Z`. Note the KML docs don't show fractional seconds, but https://www.w3.org/TR/xmlschema-2/#isoformats which they reference says ss.ssss is ok.

You should be able to call `isoformat()` on the `datetime` object to get a timestamp that I believe should be in the right format.

python/examples/extract_position_data.py
```diff
@@ -14,30 +14,82 @@
...
+"""
+
+KML_TEMPLATE_END = """\
   </Document>
 </kml>
 """
 
+KML_TEMPLATE = """\
+    <Placemark>
+      <Timestamp>%(timestamp)s</Timestamp>
```
adamshapiro0, line 84: There are 2 small (but apparently important) things here, in addition to using ISO time format like I mentioned below. The timestamp tag needs to have a capital S: `<TimeStamp>` and you need a `<when>` tag in addition like: `<TimeStamp><when>2023-10-05T21:23:06.200000Z</when></TimeStamp>`. If you do those, google earth will recognize the timestamps and provide a slider you can use to select a time range or move around in time. It's actually pretty neat.

Check out the example kml file here: https://developers.google.com/kml/documentation/time#time-stamps-and-gps-data

python/examples/extract_position_data.py
```diff
@@ -104,9 +156,12 @@
     path = os.path.join(output_dir, 'position.kml')
     logger.info("Generating '%s'." % path)
     with open(path, 'w') as f:
-        f.write(KML_TEMPLATE %
-                {'coordinates': '\n'.join(['%.8f,%.8f' % (message.lla_deg[1], message.lla_deg[0])
-                                           for message in pose_data.messages
-                                           if message.solution_type != SolutionType.Invalid])})
+        f.write(KML_TEMPLATE_START)
+        for pose in pose_data.messages:
+          f.write(KML_TEMPLATE %
```
adamshapiro0, line 161: I'd recommend we leave a note here for clarity:
> IMPORTANT: KML heights are specified in MSL, so we convert the ellipsoid heights to orthometric below using the reported geoid undulation (geoid height). Undulation values come from a geoid model, and are not typically precise. When analyzing position performance compared with another device, we strongly recommend that you do the performance using ellipsoid heights. When comparing in MSL, if the geoid models used by the two devices are not exactly the same, the heights may differ but multiple meters.

python/examples/extract_position_data.py
```diff
@@ -104,9 +156,12 @@
...
+                  {'timestamp': '\n'.join([str(pose.gps_time.as_utc())]),
+                  'solution_type':'\n'.join([str(int(pose.solution_type))]),
```
adamshapiro0, line 163: Change to %d above instead of %s and then you can remove the `str()` here

adamshapiro0 (approved):
Looks good but fix up the timestamp stuff and then test by loading a log (that moves some distance) in Google Earth. You should be able to control the time slider if it's all working correctly.

python/examples/extract_position_data.py
wbrannon: The `isoformat()` call ends up including a `00+00` at the end, and also doesn't include the desired `Z` a the end:
`2023-10-05T21:22:53.900000+00:00`

An alternative is to go with this call, which puts it in the format that you put:`.strftime('%Y-%m-%dT%H:%M:%S.%fZ')`

Tested both ways and they both work to incorporate the time slider.

adamshapiro0: Does it work with the +00:00 instead of Z? They effectively mean the same thing. If it works, I'd use the builtin function. If not, the strftime call is fine.

PR #277, opened 2023-10-13
Improvements for event table printout.

adamshapiro0:
# Changes
- Include name in reset event mask printout if a recognized mask (e.g. cold start)

# Fixes
- Fixed missing column header in event table if there are no software reset events present


wbrannon (approved):
Looks good.

PR #280, opened 2023-10-19
Clarified wheel tick output interval parameter description.

adamshapiro0:


src/point_one/fusion_engine/messages/configuration.h
```diff
@@ -1061,14 +1061,21 @@ struct P1_ALIGNAS(4) WheelConfig {
   uint8_t reserved1[1] = {0};
 
   /**
-   * The nominal rate at which wheel speed measurements will be provided (in
-   * seconds).
+   * The rate at which wheel speed/tick measurements will be sent to the device
+   * (in seconds).
```
anathan, line 1065: We should also include a comment as to why this NAN (and should be left as such)

anathan (approved):
minor addition please.

src/point_one/fusion_engine/messages/configuration.h
adamshapiro0: This parameter _is_ required. Only the output interval is derived automatically. I can change this text to clarify if you have a suggested way to make that clear?

PR #281, opened 2023-10-24
Allow applications to be run as scripts if desired.

adamshapiro0:
This allows users to run:
```
python fusion_engine_client/applications/p1_display.py
# or
./fusion_engine_client/applications/p1_display.py
```

in addition to the normally recommended:
```
p1_display
```

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -75,8 +75,9 @@ FusionEngine message specification.
           time-aligning FusionEngine data captured in a `*.p1log` file
       - `applications` - Applications and processing tools -- see [Applications](#applications) for detailed
         descriptions of each of these tools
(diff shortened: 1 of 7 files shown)


axlan (approved):
LGTM

anathan (approved):
would be nice to avoid the duplication of that code block with a one liner somehow, but this is fine for now.

adamshapiro0:
Done - I added a `enable_relative_imports()` helper function. Still requires 3 lines to call it, but it's slightly cleaner. You can now run applications as:
- `p1_display`
- `python /path/to/p1_display.py` (can be absolute or relative path, and doesn't have to be run from within the repo)
- `/path/to/p1_display.py` (same)

PR #283, opened 2023-10-31
Add new entries to `VehicleModel`

wbrannon:


python/fusion_engine_client/messages/configuration.py
```diff
@@ -105,7 +105,12 @@ class VehicleModel(IntEnum):
 
     LINCOLN_MKZ = 180,
 
-    BMW_7 = 200
+    BMW_7 = 200,
+    BMW_MOTORRAD = 201,
```
adamshapiro0, line 109: While you're in here, go ahead and remove all the commas in this class. Don't need those in Python.

adamshapiro0 (approved):
LGTM

PR #286, opened 2023-11-13
Added wheel speed input support to Python and resolved plotting issues.

adamshapiro0:
# New Features
- Added binary display options to `p1_print` (`--format={binary, pretty-binary, oneline-binary}`)
- Added missing `WheelSpeedInput` and `VehicleSpeedInput` Python classes

# Changes
- Separated wheel and vehicle speed/tick plotting to simplify code (only one should be present at a time)

# Fixes
- Clarified P1 vs system time in `MeasurementDetails` printout
- Disregard user P1 time values in wheel/vehicle speed input messages
- Fixed crash in `binary_message_decode` example when interpreting a `SetConfig` message


python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -1697,11 +1676,11 @@ def _plot_vehicle_data(data, time_source, is_raw=False, show_gear=False, style=N
                                           mode='markers', marker={'color': color}),
                              interval_y_axis, 1)
 
-        # Plot the vehicle speed data. If we have both corrected and uncorrected (raw) data, plot them both.
-        _plot_vehicle_data(vehicle_data, vehicle_time_source,
-                           is_raw=vehicle_measurement_type == raw_vehicle_measurement_type, show_gear=True)
-        if vehicle_measurement_type != raw_vehicle_measurement_type:
-            _plot_vehicle_data(raw_vehicle_data, vehicle_time_source, is_raw=True, show_gear=False)
+        # Plot the data. If we have both corrected and uncorrected (raw) data, plot them both.
```
axlan, line 1679: It took me a lot of reading to sort of understand this, but this comment doesn't seem accurate. If I'm understanding correctly, what this code does is:
`# Plot the data. If this data type has a separate corrected and uncorrected (raw) data message, plot them both.`

axlan: As written it seems to imply it's referring to whether both types are present in the log.

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -1335,59 +1335,57 @@ def plot_wheel_data(self):
...
+        self._plot_wheel_ticks_or_speeds(source='wheel', type='tick')
+        self._plot_wheel_ticks_or_speeds(source='vehicle', type='speed')
+        self._plot_wheel_ticks_or_speeds(source='vehicle', type='tick')
 
-    def _plot_wheel_ticks_or_speeds(self, type):
+    def _plot_wheel_ticks_or_speeds(self, source, type):
         """!
         @brief Plot wheel speed or tick data.
         """
         # Read the data. Try to determine which type of wheel output is present in the log (if any).
```
axlan, line 1347: This really needs a table to understand what the intent of this is. If I understand, this is to support a few different things:
1. This function is meant to be reused between ticks/speed
2. ticks and speed can be for either the wheels or the vehicle
3. there are some different messages that may be used for backwards compatibility
4. some messages support a separate raw and corrected version 
5. there are some different messages that may be used based on how the input was sent to the device in realtime
6. messages may or may not be enabled for the log

Maybe referencing this list of considerations, or a representative example would make this a bit easier to parse.

axlan (approved):
Adding some clarification on the _plot_wheel_ticks_or_speeds would be good, by overall seems fine.

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -1335,59 +1335,57 @@ def plot_wheel_data(self):
...
                                                                        DeprecatedVehicleSpeedMeasurement])
 
-            raw_wheel_measurement_type = self._auto_detect_message_type([RawWheelSpeedOutput])
-            raw_vehicle_measurement_type = self._auto_detect_message_type([RawVehicleSpeedOutput])
+            if source == 'wheel':
+                raw_measurement_type = self._auto_detect_message_type([RawWheelSpeedOutput, WheelSpeedInput])
+            else:
+                raw_measurement_type = self._auto_detect_message_type([RawVehicleSpeedOutput, VehicleSpeedInput])
+
+            if measurement_type is None:
```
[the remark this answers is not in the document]
adamshapiro0: Yeah admittedly this is a big odd/confusing. We originally did that to simplify stuff in a few places since we use the measurement type to set plot titles and such. It might be nice to just separate them completely, just need to do it carefully. I'll see how hard that would be.

In the second case with ticks, there's no such thing as corrected ticks so basically we always only have raw.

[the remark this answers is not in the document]
adamshapiro0: Ok, wasn't so bad. I separated everything so we never do `measurement_type = raw_measurement_type`.

PR #288, opened 2023-12-13
Improved early detection of invalid syncs in `FusionEngineDecoder`.

adamshapiro0:
# Changes
- Added C++ `message_decode` example app using `FusionEngineFramer` class, and renamed previous app to `raw_message_decode`

# Fixes
- Use the header reserved bytes to detect invalid syncs early when the (bogus) payload length is very large but smaller than the max packet size


src/point_one/fusion_engine/parsers/fusion_engine_framer.cc
```diff
@@ -253,37 +253,67 @@ int32_t FusionEngineFramer::OnByte(bool quiet) {
...
-          crc_check_needed = true;
+              << header->message_type << ", seq=" << header->sequence_number
+              << ", payload_size=" << header->payload_size_bytes
+              << " B, message_size=" << current_message_size_ << " B]";
+
+      // Check for size overflow. We don't currently expect to have _extremely_
+      // large packets, so this should never happen. If it does, assume this is
+      // not a valid message, and instead the sync pattern likely showed up at
+      // random in the data stream.
+      if (current_message_size_ < header->payload_size_bytes) {
```
axlan, line 268: This check appears to be wrong. Shouldn't it be `current_message_size_ < capacity_bytes_ `?

adamshapiro0: No, this one is intentional. It's checking for overflow of the uint32. The check for capacity is done below.

src/point_one/fusion_engine/parsers/fusion_engine_framer.cc
```diff
@@ -253,37 +253,67 @@ int32_t FusionEngineFramer::OnByte(bool quiet) {
...
+      // bogus header from an invalid sync has a very large payload size that
+      // happens to still be smaller than the buffer size.
+      else if (header->reserved[0] != 0 || header->reserved[1] != 0) {
+        if (quiet) {
+          VLOG(2) << "Reserved bytes nonzero. Dropping suspected invalid sync. "
+                     "[size="
+                  << current_message_size_
+                  << " B (payload=" << header->payload_size_bytes << " B)]";
+        } else {
+          LOG(WARNING) << "Message too large for buffer. [size="
```
axlan, line 299: Copy/paste needs to have description updated.

src/point_one/fusion_engine/parsers/fusion_engine_framer.cc
axlan: Oh I see this is to check for an u32 overflow. Never mind.

axlan: Though to be fair this seems like this would be simpler to have a MAX_SIZE you're checking that could be MAX_INTEGER-sizeof(MessageHeader) if you really wanted. I guess I'm not sure why you rewrote this check since the previous formulation was overflow safe and also covers the capacity check.

axlan (approved):
A couple minor comments but LGTM.

src/point_one/fusion_engine/parsers/fusion_engine_framer.cc
adamshapiro0: Yeah, I changed it mainly to make things more consistent between this and the Python code, which doesn't use uint32. It adds more lines of code but I think it also makes it a little more clear since the old one was really checking for 2 independent conditions subtly (though with a note). I'll definitely call out that this is about uint32 rollover.

PR #295, opened 2024-02-01
Fixed handling of pose messages without GPS time when generating a KML file.

adamshapiro0:
# New Features
- Added option to return `Timestamp` objects from `convert_to_*_time()` instead of just floating point values

# Fixes
- Fixed KML generation if any of the pose messages do not have GPS time, particularly after hot starting indoors


wbrannon (approved):
Looks good.

PR #300, opened 2024-03-14
Copy include_header.js automatically in Doxygen build.

adamshapiro0:
This way you can run doxygen manually on the command line without having to copy this file in separately.


axlan (approved):
LGTM

PR #305, opened 2024-04-12
Use input file prefix for output filenames in extract_* scripts.

adamshapiro0:
# Changes
- Use input file prefix for output filenames in extract_* scripts
- Added `--display-format` option to example client apps matching `p1_print`


axlan (approved):
LGTM

PR #308, opened 2024-05-03
Restored std::function callback support in FusionEngineFramer.

adamshapiro0:
# Changes
- Restored `std::function` callback support in `FusionEngineFramer`, accidentally removed in 1.22.0
- Modified C-style callback API to take a `void*` context variable as an argument


axlan (approved):
LGTM

PR #309, opened 2024-05-08
`HeadingOutput` Python and documentation cleanup.

adamshapiro0:
# New Features
- Added option to return sync bytes from `MessageHeader.unpack()` if requested

# Changes
- Removed deprecated `disable_index_generation` argument

# Fixes
- Force `offset` to 0 in all `pack()` functions if allocating a byte buffer internally
- Cleaned up `HeadingOutput` pack/unpack functions and documentation
- Added heading value to `repr(HeadingOutput)` string
- Corrected corrected heading solution type in hover text
- Enable Plotly mouse hover mode selection on heading plot

### DIFF
--- a/src/point_one/fusion_engine/messages/measurements.h
+++ b/src/point_one/fusion_engine/messages/measurements.h
@@ -1134,8 +1134,8 @@ struct P1_ALIGNAS(4) DeprecatedVehicleSpeedMeasurement : public MessagePayload {
  *        MessageType::RAW_HEADING_OUTPUT, version 1.0).
  * @ingroup measurement_messages
  *
- * This message contains raw heading sensor measurements that have not been
- * corrected for mounting angle biases.
+ * This message is an output from the device contaning raw heading sensor
+ * measurements that have not been corrected for mounting angle biases.
  *
  * See also @ref HeadingOutput.
  */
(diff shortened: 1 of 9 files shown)


src/point_one/fusion_engine/messages/measurements.h
```diff
@@ -1130,17 +1130,17 @@ struct P1_ALIGNAS(4) DeprecatedVehicleSpeedMeasurement : public MessagePayload {
...
- * @brief Raw (uncorrected) heading sensor measurement output (@ref
- *        MessageType::RAW_HEADING_OUTPUT, version 1.0).
+ * @brief Heading sensor measurement output with heading bias corrections
+ *        applied (@ref MessageType::HEADING_OUTPUT, version 1.0).
  * @ingroup measurement_messages
  *
- * This message contains raw heading sensor measurements that have not been
- * corrected for mounting angle biases.
+ * This message is an output from the device contaning heading sensor
+ * measurements that have not been corrected for horizontal and vertical biases.
```
axlan, line 1138: Maybe I'm misunderstanding this, but "measurements that have not been corrected for horizontal and vertical biases" seems to contradict "Heading sensor measurement output with heading bias corrections applied".

axlan (approved):
LGTM, with minor comment comment.

src/point_one/fusion_engine/messages/measurements.h
adamshapiro0: Oops good catch. Copy paste error. Changed to:
```
 * This message is an output from the device contaning heading sensor
 * measurements after applying horizontal and vertical bias corrections to
 * account for the orientation of the primary and secondary GNSS antennas.
 * ```

PR #311, opened 2024-05-09
Add `STARTUP_REQUEST` as command message type

wbrannon:


src/point_one/fusion_engine/messages/defs.h
```diff
@@ -295,6 +295,7 @@ P1_CONSTEXPR_FUNC bool IsCommand(MessageType message_type) {
   switch (message_type) {
     case MessageType::MESSAGE_REQUEST:
     case MessageType::RESET_REQUEST:
+    case MessageType::STARTUP_REQUEST:
```
axlan, line 298: It's a little bit annoying, but to avoid this problem in the future, you should probably restructure this switch to not have a `default`. If you explicitly handle each case in the switch, the compiler will give a warning if we add a new type and don't update this function. For extra safety you can still return false outside the switch.

axlan (approved):
LGTM, but you should restructure the switch to give a warning if we add a new type without updating this function.

PR #313, opened 2024-05-17
Added UNIX domain socket configuration settings.

adamshapiro0:
# New Features
- Added UNIX domain socket configuration settings

### DIFF
--- a/python/fusion_engine_client/messages/configuration.py
+++ b/python/fusion_engine_client/messages/configuration.py
@@ -54,6 +54,7 @@ class InterfaceConfigType(IntEnum):
   REMOTE_ADDRESS = 3
   PORT = 4
   ENABLED = 5
+  DIRECTION = 6
 
 
 class Direction(IntEnum):
@@ -163,6 +164,12 @@ class TransportType(IntEnum):
     CURRENT = 254
     ## Set/get the configuration for the all I/O interfaces.
     ALL = 255
+
+
+class TransportDirection(IntEnum):
+    INVALID = 0
+    SERVER = 1
+    CLIENT = 2
 
 
 class UpdateAction(IntEnum):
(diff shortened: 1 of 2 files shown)


src/point_one/fusion_engine/messages/configuration.h
```diff
@@ -398,13 +402,31 @@ enum class InterfaceConfigType : uint8_t {
   /**
    * Enable/disable the interface.
    *
-   * Valid for:
-   * - @ref TransportType::TCP_CLIENT
-   * - @ref TransportType::UDP_CLIENT
+   * Valid for all @ref TransportType%s.
```
axlan, line 405: `%` = `'` ?

python/fusion_engine_client/messages/configuration.py
```diff
@@ -54,6 +54,7 @@ class InterfaceConfigType(IntEnum):
   REMOTE_ADDRESS = 3
   PORT = 4
   ENABLED = 5
+  DIRECTION = 6
```
axlan, line 57: You're missing `SOCKET_TYPE `

python/fusion_engine_client/messages/configuration.py
```diff
@@ -907,34 +933,55 @@ class HeadingBias(_conf_gen.HeadingBias):
...
 
+
+@_conf_gen.create_interface_config_class(InterfaceConfigType.DIRECTION, _conf_gen.TransportDirectionConstruct)
+class InterfaceDirectionConfig(_conf_gen.TransportDirectionVal):
+    """!
+    @brief Interface transport direction (client/server) configuration settings.
+    """
+    pass
+
+
```
axlan, line 976: This should be SOCKET_TYPE and the comment should be updated.

axlan (approved):
You're missing a few parts in the Python, but otherwise LGTM.

src/point_one/fusion_engine/messages/configuration.h
adamshapiro0: No, this is doxy magic so you get a link to the type with an s after it like `TransportType`s. If you don't add the % you'll get a doxy warning that it doesn't know the class `TransportTypes`. I guess I could just change this to `all @ref TransportType values` maybe to avoid confusion.

PR #314, opened 2024-05-17
Merged TCP and UDP client/server configurations.

adamshapiro0:
# Changes
- Merged client/server configurations into one entry each for TCP and UDP
- Renamed `WEBSOCKET_SERVER` to `WEBSOCKET`


axlan (approved):
LGTM

PR #317, opened 2024-05-22
Add handling of source identifier to Python tools.

wbrannon:
# New Features
- `p1_display.py` and `p1_print.py` now accept `--source-id` as command line argument and use accordingly
- Added `MixedLogReader` and `DataLoader` options to filter data by source ID

### DIFF
--- a/python/fusion_engine_client/analysis/analyzer.py
+++ b/python/fusion_engine_client/analysis/analyzer.py
@@ -108,7 +108,7 @@ class Analyzer(object):
                  output_dir: str = None, prefix: str = '',
                  time_range: TimeRange = None, max_messages: int = None,
                  time_axis: str = 'relative',
-                 truncate_long_logs: bool = True):
+                 truncate_long_logs: bool = True, source_id: List[int]=[0]):
         """!
         @brief Create an analyzer for the specified log.
 
@@ -128,7 +128,7 @@ class Analyzer(object):
                @ref LONG_LOG_DURATION_SEC).
         """
         if isinstance(file, str):
-            self.reader = DataLoader(file, ignore_index=ignore_index)
+            self.reader = DataLoader(file, ignore_index=ignore_index, source_id=source_id)
         else:
             self.reader = file
 
@@ -141,6 +141,18 @@ class Analyzer(object):
             'show_progress': True,
             'return_numpy': True
         }
+
+        available_source_ids = self.reader.get_available_source_ids()
+        source_id_set = set(source_id)
+        available_source_id_set = set(available_source_ids)
+        if available_source_id_set != source_id_set:
+            unavailable_source_ids = list(source_id_set.difference(available_source_id_set))
+            _logger.warning('Not all source IDs requested are available. Cannot extract the following source '
+                            'IDs: {}'.format(unavailable_source_ids))
+            source_id = list(source_id_set.intersection(available_source_id_set))
+            _logger.warning('Will extract the following available requested source IDs: {}'.format(source_id))
+
+        self.source_id = source_id
 
         if time_axis in ('relative', 'rel'):
             self.time_axis = 'relative'
--- a/python/fusion_engine_client/analysis/analyzer.py
+++ b/python/fusion_engine_client/analysis/analyzer.py
@@ -327,8 +327,7 @@ class Analyzer(object):
                                  2, 1)
 
         # Read system timestamps from event notifications, if present.
-        result = self.reader.read(message_types=[EventNotificationMessage], source_ids=self.source_ids,
-                                  **self.params)
+        result = self.reader.read(message_types=[EventNotificationMessage], **self.params)
         event_data = result[EventNotificationMessage.MESSAGE_TYPE]
 
         system_time_sec = None
@@ -362,8 +361,7 @@ class Analyzer(object):
             return
 
         # Find reset events.
-        result = self.reader.read(message_types=[EventNotificationMessage], source_ids=self.source_ids,
-                                  return_message_index=True, **self.params)
+        result = self.reader.read(message_types=[EventNotificationMessage], return_message_index=True, **self.params)
         event_data = result[EventNotificationMessage.MESSAGE_TYPE]
 
         reset_idx = event_data.event_type == EventType.RESET
@@ -597,8 +595,7 @@ class Analyzer(object):
             return
 
         # Read the pose data.
-        result = self.reader.read(message_types=[CalibrationStatus], source_ids=self.source_ids,
-                                  **self.params)
+        result = self.reader.read(message_types=[CalibrationStatus], **self.params)
         cal_data = result[CalibrationStatus.MESSAGE_TYPE]
 
         if len(cal_data.p1_time) == 0:
@@ -861,7 +858,7 @@ class Analyzer(object):
             return
 
         # Read the pose data.
-        result = self.reader.read(message_types=[RelativeENUPositionMessage], source_ids=self.source_ids, **self.params)
+        result = self.reader.read(message_types=[RelativeENUPositionMessage], **self.params)
         relative_position_data = result[RelativeENUPositionMessage.MESSAGE_TYPE]
 
         if len(relative_position_data.p1_time) == 0:
@@ -980,8 +977,7 @@ class Analyzer(object):
 
     def plot_gnss_skyplot(self, decimate=True):
         # Read the satellite data.
-        result = self.reader.read(message_types=[GNSSSatelliteMessage], source_ids=self.source_ids,
-                                  **self.params)
+        result = self.reader.read(message_types=[GNSSSatelliteMessage], **self.params)
         data = result[GNSSSatelliteMessage.MESSAGE_TYPE]
 
         if len(data.p1_time) == 0:
@@ -1077,8 +1073,7 @@ class Analyzer(object):
     def plot_gnss_cn0(self):
         # The legacy GNSSSatelliteMessage contains data per satellite, not per signal. The plotted C/N0 values will
         # reflect the L1 signal, unless L1 is not being tracked.
-        result = self.reader.read(message_types=[GNSSSatelliteMessage], source_ids=self.source_ids,
-                                  **self.params)
+        result = self.reader.read(message_types=[GNSSSatelliteMessage], **self.params)
         data = result[GNSSSatelliteMessage.MESSAGE_TYPE]
 
         if len(data.p1_time) == 0:
@@ -1140,8 +1135,7 @@ class Analyzer(object):
         figure_title = "GNSS Signal Status"
 
         # Read the satellite data.
-        result = self.reader.read(message_types=[GNSSSatelliteMessage], source_ids=self.source_ids,
-                                  **self.params)
+        result = self.reader.read(message_types=[GNSSSatelliteMessage], **self.params)
         data = result[GNSSSatelliteMessage.MESSAGE_TYPE]
         is_legacy_message = True
 
@@ -1274,7 +1268,7 @@ Gold=Float, Green=Integer (Not Fixed), Blue=Integer (Fixed, Float Solution Type)
 
         This includes geometric, position, horizontal, and vertical DOP.
         """
-        result = self.reader.read(message_types=[GNSSInfoMessage], source_ids=self.source_ids, **self.params)
+        result = self.reader.read(message_types=[GNSSInfoMessage], **self.params)
         data = result[GNSSInfoMessage.MESSAGE_TYPE]
 
         if len(data.p1_time) == 0:
@@ -1311,7 +1305,7 @@ Gold=Float, Green=Integer (Not Fixed), Blue=Integer (Fixed, Float Solution Type)
         """!
         @brief Plot GNSS corrections status (baseline distance, age, etc.).
         """
-        result = self.reader.read(message_types=[GNSSInfoMessage], source_ids=self.source_ids, **self.params)
+        result = self.reader.read(message_types=[GNSSInfoMessage], **self.params)
         data = result[GNSSInfoMessage.MESSAGE_TYPE]
 
         if len(data.p1_time) == 0:
@@ -1422,8 +1416,7 @@ Gold=Float, Green=Integer (Not Fixed), Blue=Integer (Fixed, Float Solution Type)
         if self.long_log_detected and self.truncate_data:
             params = copy.deepcopy(self.params)
             params['max_messages'] = 2
-            result = self.reader.read(message_types=any_measurement_type, remove_nan_times=False,
-                                      source_ids=self.source_ids, **params)
+            result = self.reader.read(message_types=any_measurement_type, remove_nan_times=False, **params)
             data = result[any_measurement_type.MESSAGE_TYPE]
             if len(data.measurement_time) == 2:
                 dt_sec = data.measurement_time[1] - data.measurement_time[0]
@@ -1436,7 +1429,7 @@ Gold=Float, Green=Integer (Not Fixed), Blue=Integer (Fixed, Float Solution Type)
 
         # Read the data.
         result = self.reader.read(message_types=[measurement_type, raw_measurement_type],
-                                  remove_nan_times=False, source_ids=self.source_ids, **self.params)
+                                  remove_nan_times=False, **self.params)
 
         def _extract_data(measurement_type):
             if measurement_type is not None:
@@ -1736,7 +1729,7 @@ Gold=Float, Green=Integer (Not Fixed), Blue=Integer (Fixed, Float Solution Type)
         if self.truncate_data:
             params = copy.deepcopy(self.params)
             params['max_messages'] = 2
-            result = self.reader.read(message_types=[message_cls], source_ids=self.source_ids, **params)
+            result = self.reader.read(message_types=[message_cls], **params)
             data = result[message_cls.MESSAGE_TYPE]
             if len(data.p1_time) == 2:
                 dt_sec = data.p1_time[1] - data.p1_time[0]
@@ -1748,7 +1741,7 @@ Gold=Float, Green=Integer (Not Fixed), Blue=Integer (Fixed, Float Solution Type)
                     return
 
         # Read the data.
-        result = self.reader.read(message_types=[message_cls], source_ids=self.source_ids, **self.params)
+        result = self.reader.read(message_types=[message_cls], **self.params)
         data = result[message_cls.MESSAGE_TYPE]
 
         if len(data.p1_time) == 0:
@@ -1808,8 +1801,7 @@ Gold=Float, Green=Integer (Not Fixed), Blue=Integer (Fixed, Float Solution Type)
             return
 
         # Read the heading measurement data.
-        result = self.reader.read(message_types=[RawHeadingOutput, HeadingOutput], source_ids=self.source_ids,
-                                  **self.params)
+        result = self.reader.read(message_types=[RawHeadingOutput, HeadingOutput], **self.params)
         raw_heading_data = result[RawHeadingOutput.MESSAGE_TYPE]
         heading_data = result[HeadingOutput.MESSAGE_TYPE]
 
@@ -2056,8 +2048,7 @@ Gold=Float, Green=Integer (Not Fixed), Blue=Integer (Fixed, Float Solution Type)
             return
 
         # Read the data.
-        result = self.reader.read(message_types=[SystemStatusMessage], remove_nan_times=False,
-                                  source_ids=self.source_ids, **self.params)
+        result = self.reader.read(message_types=[SystemStatusMessage], remove_nan_times=False, **self.params)
         data = result[SystemStatusMessage.MESSAGE_TYPE]
 
         if len(data.p1_time) == 0:
@@ -2092,8 +2083,7 @@ Gold=Float, Green=Integer (Not Fixed), Blue=Integer (Fixed, Float Solution Type)
 
         # Read the data.
         data = self.reader.read(message_types={MessageType.EVENT_NOTIFICATION} | COMMAND_MESSAGES | RESPONSE_MESSAGES,
-                                remove_nan_times=False, return_in_order=True, return_bytes=True,
-                                source_ids=self.source_ids, **self.params)
+                                remove_nan_times=False, return_in_order=True, return_bytes=True, **self.params)
 
         if len(data.messages) == 0:
             self.logger.info('No event notification data available.')
@@ -2293,15 +2283,13 @@ Gold=Float, Green=Integer (Not Fixed), Blue=Integer (Fixed, Float Solution Type)
         params['max_messages'] = 1
         params['return_in_order'] = True
 
-        result = self.reader.read(message_types=None, require_p1_time=True, source_ids=self.source_ids,
-                                  **params)
+        result = self.reader.read(message_types=None, require_p1_time=True, **params)
         if len(result.messages) > 0:
             processed_t0 = result.messages[0].get_p1_time()
         else:
             processed_t0 = Timestamp()
 
-        result = self.reader.read(message_types=None, require_system_time=True, source_ids=self.source_ids,
-                                  **params)
+        result = self.reader.read(message_types=None, require_system_time=True, **params)
         if len(result.messages) > 0:
             processed_system_t0 = result.messages[0].get_system_time_sec()
         else:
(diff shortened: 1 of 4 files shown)


python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -911,14 +899,44 @@ def _plot_data(name, idx, marker_style=None):
...
 
-        for type, info in _SOLUTION_TYPE_MAP.items():
-            _plot_data(info.name, solution_type == type, marker_style=info.style)
+        # Read the pose data.
+        for source_id in self.source_ids:
+            result = self.reader.read(message_types=[PoseMessage], source_ids=[source_id], **self.params)
+            pose_data = result[PoseMessage.MESSAGE_TYPE]
+
+            if len(pose_data.p1_time) == 0:
+                self.logger.info('No pose data available. Skipping map display.')
```
ryancdougherty, line 920: Add source ID to this log.

python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -556,6 +596,29 @@ def filter_out_invalid_p1_times(self, clear_existing: bool = False):
 
         return self
 
+    # Move to mixed log reader
```
ryancdougherty, line 599: Comment no longer applicable? (this is in mixed_log_reader.py)

python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -107,6 +111,18 @@ def __init__(self, input_file, warn_on_gaps: bool = False, show_progress: bool =
         self._original_index = fast_indexer.fast_generate_index(input_path, force_reindex=ignore_index,
                                                                 save_index=save_index, max_bytes=max_bytes)
         self.next_index_elem = 0
+        self.index = self._original_index
+        self.filtered_message_types = False
+
+        # Now, apply filters -> apply pose message filter, then look at available message types
+        # self.populate_xx() (builds list of source_ids) -> self.filter_in_place(source_id) -> applies pose message filtering
+            # get list of available source IDs
```
ryancdougherty, line 119: indent is off

python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -107,6 +111,18 @@ def __init__(self, input_file, warn_on_gaps: bool = False, show_progress: bool =
         self._original_index = fast_indexer.fast_generate_index(input_path, force_reindex=ignore_index,
                                                                 save_index=save_index, max_bytes=max_bytes)
         self.next_index_elem = 0
+        self.index = self._original_index
+        self.filtered_message_types = False
+
+        # Now, apply filters -> apply pose message filter, then look at available message types
```
ryancdougherty, line 117: What's up with this comment? Does this match the style of other comments in this repo? What does it mean?

python/fusion_engine_client/applications/p1_print.py
```diff
@@ -123,6 +125,9 @@ def main():
         '--skip', type=int, default=0,
         help="Skip the first N messages in the log. If --message-type is specified, only count messages matching the "
              "specified type(s).")
+    parser.add_argument(
+        '--source-identifier', '--source-id', nargs='*', type=int, default=0,
+        help="Print the Fusion Engine messages characterized by the listed source identifier(s).")
```
ryancdougherty, line 130: "characterized by" to "with" perhaps?

ryancdougherty: Does this work for a list? What format should the user enter for a list?

python/fusion_engine_client/analysis/data_loader.py
```diff
@@ -172,7 +174,10 @@ def open(self, path, save_index=True, ignore_index=False):
         self.close()
 
         self.reader = MixedLogReader(input_file=path, save_index=save_index, ignore_index=ignore_index,
-                                     return_bytes=True, return_message_index=True)
+                                     return_bytes=True, return_message_index=True, source_ids=self.source_ids)
+
+        # Narrow down source IDs to the ones that are requested AND are available.
+        self.source_ids = self.reader.requested_source_ids
```
ryancdougherty, line 180: Code and comment don't seem to match. Is that because the code is non-obvious? or because the comment is incorrect?

Looking a little deeper, it looks like the code is not obvious. On instantiation, the `reader` is going to populate `requested_source_ids` not purely from a request but from the intersection of a request and what is actually available. That might be worth clarifying. If you can think of a way to clarify this behavior by way of clever naming, rather than a comment, that is preferable.

python/fusion_engine_client/analysis/data_loader.py
```diff
@@ -468,6 +475,8 @@ def _read(self,
             except StopIteration:
                 break
 
+            # if header.source_identifier not in self.requested_source_id:
+            #     continue
```
ryancdougherty, line 479: remove commented code.

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -911,14 +899,44 @@ def _plot_data(name, idx, marker_style=None):
...
+                return
+
+            # Remove invalid solutions.
+            valid_idx = np.logical_and(~np.isnan(pose_data.p1_time), pose_data.solution_type != SolutionType.Invalid)
+            if not np.any(valid_idx):
+                self.logger.info('No valid position solutions detected.')
```
ryancdougherty, line 926: Add source ID to this log as well.

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -2528,6 +2556,10 @@ def main():
              "Options include:%s" %
              ''.join(['\n- %s' % f[5:] for f in plot_function_names]))
 
+    plot_group.add_argument(
+        '--source-identifier', '--source-id', nargs='*', type=int,
+        help="Plot the Fusion Engine messages characterized by the listed source identifier(s).")
```
ryancdougherty, line 2561: "characterized by" to "with" or perhaps "from"

ryancdougherty (approved):
I don't see any obvious problems, but I'm not too familiar with this repo.

python/fusion_engine_client/applications/p1_print.py
adamshapiro0: I suggest using `CSVAction` for arguments that can be specified more than once. See the use in `analyzer.py`.

python/fusion_engine_client/analysis/data_loader.py
```diff
@@ -135,7 +135,7 @@ class DataLoader(object):
 
     logger = logging.getLogger('point_one.fusion_engine.analysis.data_loader')
 
-    def __init__(self, path=None, save_index=True, ignore_index=False):
+    def __init__(self, path=None, save_index=True, ignore_index=False, source_ids: List[int]=[0]):
```
adamshapiro0, line 138: This should be optional and should default to `None`. Do not assume that 0 is present, or is the only ID that the user wants by default.

python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -76,6 +77,9 @@ def __init__(self, input_file, warn_on_gaps: bool = False, show_progress: bool =
             if len(self.message_types) == 0:
                 self.message_types = None
 
+        # The source IDs requested by the user that are available.
+        self.requested_source_ids = source_ids
```
adamshapiro0, line 81: Convert the list to a set here so you can treat it as a set later for taking intersections

python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -107,6 +111,18 @@ def __init__(self, input_file, warn_on_gaps: bool = False, show_progress: bool =
         self._original_index = fast_indexer.fast_generate_index(input_path, force_reindex=ignore_index,
                                                                 save_index=save_index, max_bytes=max_bytes)
         self.next_index_elem = 0
+        self.index = self._original_index
+        self.filtered_message_types = False
+
+        # Now, apply filters -> apply pose message filter, then look at available message types
+        # self.populate_xx() (builds list of source_ids) -> self.filter_in_place(source_id) -> applies pose message filtering
+            # get list of available source IDs
+        self.get_available_source_ids()
```
adamshapiro0, line 120: You don't define `self.available_source_ids` in the constructor. You need to do that before calling `get_available_source_ids()` -- all member variables should always be declared in the constructor for clarity, even though technically Python lets you do it later.

python/fusion_engine_client/parsers/mixed_log_reader.py
adamshapiro0, line 81: Also define `self.available_source_ids = None` here and add a note that it will be populated below.

python/fusion_engine_client/parsers/mixed_log_reader.py
adamshapiro0: This comment doesn't really match the code. I'm not totally sure what this sentence is saying, so it needs clarification, and there is no `populate_xx()` function?

I think this could be better said as like:

> Before restricting to the user-requested set of source IDs, first look at data in the file to determine the list of all available source IDs (within the user's time range).

This step should apply the time range filter but _not_ the message type filter, then call `get_available_source_ids()` to generate the list of IDs, which will be stored in `self.available_source_ids` (which should be a `set`).

Next, step 2 if the user specified the `source_ids` argument here (stored in `self.requested_source_ids`), I would update `self.available_source_ids` to be the intersection of the two. Later, in `read()`:
```py
if source_ids is None:
  source_ids = None self.requested_source_ids is None else self.available_source_ids
else:
  source_ids = self.available_source_ids.intersection(source_ids)
```
If `source_ids is not None` you apply it to the messages. Otherwise you can skip checking IDs altogether.

Finally, step 3 should clear the filters and reapply the time range, plus the message types (with a comment explaining it).

---

I would actually recommend you _do_ create a populate function to call here to make things a little more explicit and clear, but I would make it protected: `_populate_available_source_ids()`. Then `get_available_source_ids()` should simply return `self.available_source_ids`.

adamshapiro0: Actually no, scratch most of that. I see how `self.requested_source_ids` is being used in `filter_in_place()`, and it is consistent with time range and message types.

We should still separate populate and get, but `self.available_source_ids` should simply be the list of all available source IDs regardless of time range or anything else.

We should make the argument default above `None` though, not `[0]`, and both `self.available_source_ids` and `self.requested_source_ids` should be converted to a `set()` for efficiency and convenience.

python/fusion_engine_client/analysis/data_loader.py
adamshapiro0: Correction: remove this altogether

python/fusion_engine_client/analysis/data_loader.py
adamshapiro0: We should not be passing `source_ids` to the log reader constructor in this class. We are treating it the same way we do time range and message types, so it should be passed to `read()` only

python/fusion_engine_client/analysis/data_loader.py
```diff
@@ -282,7 +287,7 @@ def _read(self,
               return_numpy: bool = False, keep_messages: bool = False, remove_nan_times: bool = True,
               time_align: TimeAlignmentMode = TimeAlignmentMode.NONE,
               aligned_message_types: Union[list, tuple, set] = None,
-              quiet: bool = False) \
+              quiet: bool = False, source_ids: List[int] = [0]) \
```
adamshapiro0, line 290: You need to document this argument, but you should move it up with `message_types` and `time_range` (and then add the documentation in the right place). The arguments here are organized into what they do: filter the data, control the returned values, etc.

Also, this should be `Optional` and should default to `None`.

python/fusion_engine_client/analysis/data_loader.py
```diff
@@ -309,6 +314,7 @@ def _read(self,
             'return_bytes': return_bytes,
             'return_message_index': return_message_index,
             'remove_nan_times': remove_nan_times,
+            'source_ids': source_ids,
```
adamshapiro0, line 317: Convert to  set so order won't matter. Otherwise, if the user does `[0, 3]` one time and then `[3, 0]` another time the lists won't be equal and the cached data will be cleared.

python/fusion_engine_client/applications/p1_print.py
```diff
@@ -182,6 +187,11 @@ def main():
     # Parse the time range.
     time_range = TimeRange.parse(options.time, absolute=options.absolute_time)
 
+    if options.source_identifier is None:
+        source_ids = [0]
```
adamshapiro0, line 191: Please go through carefully and remove this type of thing since it shows up a lot. We should use `None` by default everywhere and display everything, not assume `[0]` is right.

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -142,6 +142,9 @@ def __init__(self,
             'return_numpy': True
         }
 
+        # For plots that only need to trace one source ID.
+        self.default_source_id = min(self.reader.source_ids)
```
adamshapiro0, line 146: Use `get_available_source_ids()` instead of accessing this member directly

adamshapiro0: Also, if the user does specify the `source_ids` argument, you should set this to the intersection of the available IDs and the ones the user requested. If the intersection is empty, there will be no available data at all and we should print an error and abort.

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -128,7 +128,7 @@ def __init__(self,
                @ref LONG_LOG_DURATION_SEC).
         """
         if isinstance(file, str):
-            self.reader = DataLoader(file, ignore_index=ignore_index)
+            self.reader = DataLoader(file, ignore_index=ignore_index, source_ids=source_id)
```
adamshapiro0, line 131: I'd suggest a few changes here:
- Change the param name above to `source_ids` for consistency with other files
- Set the default to `None` (and make the type `Optional[List[int]]`) -- the default behavior should be "load all available data", and should also not assume that 0 is an available ID (even though it typically is)
- Don't specify `source_ids` in the `DataLoader` constructor (in fact as noted in that file, remove this argument)
- Remove the whole `default_source_id` business. The intent of this argument is really for the user to specify which pose source they're interested in. If they do `--source-id=1`, there's no reason for us to drop all event notifications for example

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -2528,6 +2556,10 @@ def main():
              "Options include:%s" %
              ''.join(['\n- %s' % f[5:] for f in plot_function_names]))
 
+    plot_group.add_argument(
+        '--source-identifier', '--source-id', nargs='*', type=int,
```
adamshapiro0, line 2560: Use CSVAction

adamshapiro0:
@wbrannon FYI the Python unit tests are failing on this branch

python/fusion_engine_client/analysis/data_loader.py
wbrannon: The problem that I'm seeing with getting rid of this is that the mixed log reader needs the source IDs in its constructor, which isn't called until the data loader's `open` function.

python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -25,7 +25,8 @@ def __init__(self, input_file, warn_on_gaps: bool = False, show_progress: bool =
                  save_index: bool = True, ignore_index: bool = False, max_bytes: int = None,
                  time_range: TimeRange = None, message_types: Union[Iterable[MessageType], MessageType] = None,
                  return_header: bool = True, return_payload: bool = True,
-                 return_bytes: bool = False, return_offset: bool = False, return_message_index: bool = False):
+                 return_bytes: bool = False, return_offset: bool = False, return_message_index: bool = False,
+                 source_ids: List[int] = None):
```
adamshapiro0, line 29: Missing optional in type hint

python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -76,8 +77,16 @@ def __init__(self, input_file, warn_on_gaps: bool = False, show_progress: bool =
             if len(self.message_types) == 0:
                 self.message_types = None
 
+        # The source IDs requested by the user. If none were requested, then use all of them.
+        if source_ids is None:
+            self.requested_source_ids = None
+        else:
+            self.requested_source_ids = set(source_ids)
+        # The source IDs that are available in the log. This will be populated below when get_available_source_ids is
```
adamshapiro0, line 85: _populate_available_source_ids(), not get

python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -556,6 +596,33 @@ def filter_out_invalid_p1_times(self, clear_existing: bool = False):
 
         return self
 
+    def get_available_source_ids(self, num_messages_to_read: int = 10):
+        self._populate_available_source_ids(num_messages_to_read=num_messages_to_read)
```
adamshapiro0, line 600: This doesn't make any sense. `get_available_source_ids()` is meant to be a user-facing getter. It should just return `self.available_source_ids`.

`_populate_available_source_ids()` is totally separate and should be called by the constructor to populate the list one time and then not again.

python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -107,6 +116,14 @@ def __init__(self, input_file, warn_on_gaps: bool = False, show_progress: bool =
         self._original_index = fast_indexer.fast_generate_index(input_path, force_reindex=ignore_index,
                                                                 save_index=save_index, max_bytes=max_bytes)
         self.next_index_elem = 0
+        self.index = self._original_index
+        self.filtered_message_types = False
+
+        self.get_available_source_ids()
+        self.filter_in_place(None, source_ids=self.requested_source_ids)
+
```
adamshapiro0, line 124: I suggest this blank and the line above and grouping all of the filter in place calls together

python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -466,9 +488,27 @@ def filter_in_place(self, key, clear_existing: bool = False):
...
+            source_ids = set(source_ids)
+            # Apply source IDs that are available.
+            if self.available_source_ids != source_ids:
+                unavailable_source_ids = list(source_ids.difference(self.available_source_ids))
+                if len(unavailable_source_ids) > 0:
+                    self.logger.warning('Not all source IDs requested are available. Cannot extract the following '
+                                        'source IDs: {}'.format(unavailable_source_ids))
+                source_ids = list(source_ids.intersection(self.available_source_ids))
+                if len(source_ids) == 0:
+                    raise ValueError("Requested source ID(s) unavailable. Exiting.")
```
adamshapiro0, line 506: This message is a little confusing. First exceptions don't mean the application has to exit, that's not their decision. But also "unavailable" implies something within the class. Here this is simply an input error: when specifying a list of desired source IDs, it doesn't make sense for the list to be empty.

Something like "Requested source ID list cannot be empty." would be more clear.

python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -466,9 +488,27 @@ def filter_in_place(self, key, clear_existing: bool = False):
...
                 self.index = self._original_index
 
+        # Set requested source IDs.
+        if source_ids is not None:
+            source_ids = set(source_ids)
+            # Apply source IDs that are available.
+            if self.available_source_ids != source_ids:
+                unavailable_source_ids = list(source_ids.difference(self.available_source_ids))
+                if len(unavailable_source_ids) > 0:
+                    self.logger.warning('Not all source IDs requested are available. Cannot extract the following '
```
adamshapiro0, line 502: I don't think this should be a warning. Maybe a debug print, but this is totally normal. The user's list is "this is what I would like _if available_", not "you must give me data for all of these".

python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -556,6 +596,33 @@ def filter_out_invalid_p1_times(self, clear_existing: bool = False):
...
+        self._populate_available_source_ids(num_messages_to_read=num_messages_to_read)
+
+    def _populate_available_source_ids(self, num_messages_to_read):
+        self.available_source_ids = set()
+        # Loop over all message types and read N of each type.
+        for message_type in np.unique(self.index['type']):
+            try:
+                message_type = MessageType(message_type, raise_on_unrecognized=True)
+            except (KeyError, ValueError) as e:
+                continue
```
adamshapiro0, line 609: Instead of this, I think we should allow unrecognized types above and then treat them the same as all others below. Source ID is part of the message header, so there's no need to decode the payload to get it and unsupported types (i.e., types we don't have classes defined for and that we can't decode the payload for) should still count.

You should test this just to be safe to make sure it does actually work when you make this change. Change the pose message enum definition to a wrong value, then process a log and make sure it still handles message 10000 correctly and populates the list of source IDs.

python/fusion_engine_client/analysis/data_loader.py
```diff
@@ -282,7 +287,7 @@ def _read(self,
               return_numpy: bool = False, keep_messages: bool = False, remove_nan_times: bool = True,
               time_align: TimeAlignmentMode = TimeAlignmentMode.NONE,
               aligned_message_types: Union[list, tuple, set] = None,
-              quiet: bool = False) \
+              quiet: bool = False, source_ids: Optional[List[int]] = None) \
```
adamshapiro0, line 290: This parameter needs to be documented above. I already noted this, not sure where that comment went, but this should be moved up above so it is defined with the other filtering arguments (message types, time range), and the documentation order should match.

python/fusion_engine_client/parsers/mixed_log_reader.py
adamshapiro0: On second thought, right now we don't have a way to reset the source IDs filter when you call `filter_in_place()`. You can only clear all filters and then start over. If you pass in `None`, it gets ignored to avoid undoing one filter when you're changing another.

We should maybe treat an empty set as an explicit "allow all" instead.

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -311,7 +327,8 @@ def _calc_stats(input):
                                  2, 1)
 
         # Read system timestamps from event notifications, if present.
-        result = self.reader.read(message_types=[EventNotificationMessage], **self.params)
+        result = self.reader.read(message_types=[EventNotificationMessage], source_ids=self.source_ids,
```
adamshapiro0, line 330: Here and below, the majority of these message types don't care about source IDs and should probably just read and display all. This one for example. You should be able to simply not set `source_ids` at all for these. The default behavior should be to read all available sources.

The only read calls that need to set the source ID list I think should be pose and pose aux currently.

python/fusion_engine_client/parsers/mixed_log_reader.py
wbrannon: This is actually supposed to convey that all the source IDs requested by the user are unavailable!

adamshapiro0: Ah I see, this is after the intersection. That's ok then, but we should probably just print a warning. Reading should work gracefully like normal, so it'll basically just output nothing. Same as if you gave it a time range that didn't cover the data. It shouldn't raise an exception.

python/fusion_engine_client/parsers/mixed_log_reader.py
wbrannon: Okay this does actually give us an error; the mixed log reader doesn't work when we call `filter_in_place()` with an unsupported message type. In particular, this line gives an error because it's an "unsupported key type":
```
self.index = self.index[key]
```

python/fusion_engine_client/parsers/mixed_log_reader.py
```diff
@@ -460,14 +483,44 @@ def filter_in_place(self, key, clear_existing: bool = False):
                 # that we just read.
                 prev_offset_bytes = self.index.offset[self.next_index_elem - 1]
 
-        # If requested, clear previous filter criteria.
-        if clear_existing:
-            if self.index is None:
+        if type(clear_existing) == str:
```
adamshapiro0, line 486: Use `instanceof(clear_existing, str)` instead

adamshapiro0 (approved):
Looks good

PR #318, opened 2024-05-22
Add tool for filtering FE data in realtime.

axlan:
# New Features
- Allow unknown `MessageTypes` to be specified as args
- Tool to filter FusionEngine data over stdin

# Fixes
- Avoid error with empty `fast_indexer` results

### DIFF
--- a/python/fusion_engine_client/applications/p1_filter.py
+++ b/python/fusion_engine_client/applications/p1_filter.py
@@ -1,0 +1,103 @@
+#!/usr/bin/env -S python3 -u
+
+from datetime import datetime
+import os
+import sys
+import time
+
+# Since stdout is used for data stream, don't write any print statements to stdout.
+# Done here to avoid any log/print statements triggered by imports.
+original_stdout = sys.stdout
+sys.stdout = sys.stderr
+
+# Add the Python root directory (fusion-engine-client/python/) to the import search path to enable FusionEngine imports
+# if this application is being run directly out of the repository and is not installed as a pip package.
+root_dir = os.path.normpath(os.path.join(os.path.dirname(__file__), '..'))
+sys.path.insert(0, root_dir)
+
+from fusion_engine_client.messages import MessagePayload, message_type_by_name
+from fusion_engine_client.parsers import FusionEngineDecoder
+from fusion_engine_client.utils.argument_parser import ArgumentParser, ExtendedBooleanAction
+
+
+if __name__ == "__main__":
+    parser = ArgumentParser(description="""\
+Filter FusionEngine data coming through stdin. Examples:
+  netcat 192.168.1.138 30210 | ./p1_filter.py --blacklist -m SatelliteInfo --display > /tmp/out.p1log
+  cat /tmp/out.p1log | ./p1_filter.py -m Pose > /tmp/pose_out.p1log
+  stty -F /dev/ttyUSB0 speed 460800 cs8 -cstopb -parenb -icrnl -ixon -ixoff -opost -isig -icanon -echo && cat /dev/ttyUSB0 | ./p1_filter.py -m Pose > /tmp/pose_out.p1log
+""")
+
+    parser.add_argument(
+        '-m', '--message-type', type=str, action='append',
+        help="An list of class names corresponding with the message types to whitelist (default) or blacklist. May be specified "
+             "multiple times (-m Pose -m PoseAux), or as a comma-separated list (-m Pose,PoseAux). All matches are"
+             "case-insensitive.\n"
+             "\n"
+             "If a partial name is specified, the best match will be returned. Use the wildcard '*' to match multiple "
+             "message types.\n"
+             "\n"
+             "Supported types:\n%s" % '\n'.join(['- %s' % c for c in message_type_by_name.keys()]))
+    parser.add_argument(
+        '--blacklist', action=ExtendedBooleanAction,
+        help="By default '--message-type' is a whitelist. If 'blacklist' is set, use them as a blacklist instead.")
+    parser.add_argument(
+        '--display', action=ExtendedBooleanAction,
+        help="Periodically display messages received and forwarded on sterr.")
+    options = parser.parse_args()
+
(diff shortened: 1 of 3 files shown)


python/fusion_engine_client/messages/defs.py
```diff
@@ -453,8 +454,13 @@ def find_matching_message_types(cls, pattern: Union[str, List[str]], return_clas
         for pattern in requested_types:
             # Check if pattern is the message integer value.
             try:
-                int_val = int(pattern)
-                result.add(MessageType(int_val))
+                # Allow specifying unknown MessageType values if they are preceded by 'u'.
+                if len(pattern) > 1 and pattern[0] == 'u':
```
adamshapiro0, line 458: Is this u prefix really necessary? Any reason not to just treat any manually specified integer as "might be unknown and that's allowable"?

python/fusion_engine_client/applications/p1_filter.py
```diff
@@ -0,0 +1,103 @@
...
+from fusion_engine_client.parsers import FusionEngineDecoder
+from fusion_engine_client.utils.argument_parser import ArgumentParser, ExtendedBooleanAction
+
+
+if __name__ == "__main__":
+    parser = ArgumentParser(description="""\
+Filter FusionEngine data coming through stdin. Examples:
+  netcat 192.168.1.138 30210 | ./p1_filter.py --blacklist -m SatelliteInfo --display > /tmp/out.p1log
+  cat /tmp/out.p1log | ./p1_filter.py -m Pose > /tmp/pose_out.p1log
+  stty -F /dev/ttyUSB0 speed 460800 cs8 -cstopb -parenb -icrnl -ixon -ixoff -opost -isig -icanon -echo && cat /dev/ttyUSB0 | ./p1_filter.py -m Pose > /tmp/pose_out.p1log
```
adamshapiro0, line 28: You might want to break these longer command examples onto multiple lines at 80 chars with \ for readability. The arg parser description stuff won't wrap this automatically I don't think, so for most default terminal sizes this will be slightly hard to read as is.
```
 stty ... \
    -isig -icanon -echo && \
    cat /dev/ttyUSB0 | \
./p1_filter.py ...
```

python/fusion_engine_client/applications/p1_filter.py
```diff
@@ -0,0 +1,103 @@
...
+             "multiple times (-m Pose -m PoseAux), or as a comma-separated list (-m Pose,PoseAux). All matches are"
+             "case-insensitive.\n"
+             "\n"
+             "If a partial name is specified, the best match will be returned. Use the wildcard '*' to match multiple "
+             "message types.\n"
+             "\n"
+             "Supported types:\n%s" % '\n'.join(['- %s' % c for c in message_type_by_name.keys()]))
+    parser.add_argument(
+        '--blacklist', action=ExtendedBooleanAction,
+        help="By default '--message-type' is a whitelist. If 'blacklist' is set, use them as a blacklist instead.")
```
adamshapiro0, line 43: This whitelist and blacklist terminology isn't quite clear here and in `-m` above. It would help to be more explicit like "If specified, discard all message types specified with --message-type and output everything else. By default, all specified message types are output and all others are discarded." Or something like that.

python/fusion_engine_client/applications/p1_filter.py
```diff
@@ -0,0 +1,103 @@
...
+    parser.add_argument(
+        '--display', action=ExtendedBooleanAction,
+        help="Periodically display messages received and forwarded on sterr.")
```
adamshapiro0, line 46: sterr --> stderr?

This sentence is a little ambiguous. I think you're saying that you will display the content on the stderr, but it could be read as "the messages that were forwarded on stderr".

python/fusion_engine_client/applications/p1_filter.py
```diff
@@ -0,0 +1,103 @@
...
+                    if pass_through_message:
+                        messages_forwarded += 1
+                        bytes_forwarded += len(raw_data)
+                        original_stdout.buffer.write(raw_data)
+
+            if options.display:
+                now = datetime.now()
+                if (now - last_print_time).total_seconds() > 5.0:
+                    print('Status: [bytes_received=%d, messages_received=%d, bytes_forwarded=%d, messages_forwarded=%d, elapsed_time=%d sec]' %
+                        (bytes_received, messages_received, bytes_forwarded, messages_forwarded, (now - start_time).total_seconds()))
```
adamshapiro0, line 99: max 120 chars please

python/fusion_engine_client/applications/p1_filter.py
```diff
@@ -0,0 +1,103 @@
...
+            # This returns immediately even if 0 bytes are available.
+            received_data = sys.stdin.buffer.read(64)
+            if len(received_data) == 0:
+                time.sleep(0.1)
+            else:
+                bytes_received += len(received_data)
+                messages = decoder.on_data(received_data)
+                for (header, message, raw_data) in messages:
+                    messages_received += 1
+                    pass_through_message = (options.blacklist and header.message_type not in message_types) or (not options.blacklist and header.message_type in message_types)
```
adamshapiro0, line 89: 120

python/fusion_engine_client/applications/p1_filter.py
adamshapiro0: Also I don't think it's right? Down below you just print periodic status, you don't actually display messages at all.

adamshapiro0 (approved):
Looks good, just some minor cleanup for clarity.

python/fusion_engine_client/messages/defs.py
axlan: The thought was that 99% of the time an external customer would want to filter known messages and this would allow detecting typos.

adamshapiro0: I think in practice this will just be an annoyance more than helpful. The majority of the time someone using this tool is probably going to be specifying message types by name, not number.

PR #321, opened 2024-06-06
Skip empty files by default when searching for a p1log file.

adamshapiro0:
# Changes
- Skip empty files by default in Python `find_log_file()` function


axlan (approved):
LGTM

PR #322, opened 2024-06-12
Moved version, device ID, and event messages, and cleaned up Doxygen.

adamshapiro0:
# Changes
- Moved `VersionInfoMessage`, `DeviceIDMessage`, and `EventNotificationMessage` from `control.h` to `device.h`
- Reorganized `configuration.h` Doxygen to group components and improve navigation
- Removed commented-out arrays from variable-length message structs to avoid confusion
- Minor documentation wording clarifications

# Fixes
- Fixed overlapping pose sources in time scale plot
- Fixed speed estimate on wheel speed plot if multiple pose sources are present


adamshapiro0:
Note to self: fix `WheelTickInput` class documentation, and presumably vehicle tick too.

axlan (approved):
LGTM

src/point_one/fusion_engine/messages/measurements.h
```diff
@@ -809,10 +809,8 @@ struct P1_ALIGNAS(4) RawVehicleSpeedOutput : public MessagePayload {
  * @ingroup measurement_messages
  *
  * This message is an input to the device, used to convey the wheel encoder tick
- * counts for one or more wheels, received through software (e.g., vehicle CAN
- * bus) or captured in hardware from external voltage pulses. The number and
- * type of wheels expected, and the interpretation of the tick count values,
- * varies by vehicle.
+ * counts for one or more wheels. The number and type of wheels expected, and
+ * the interpretation of the tick count values, varies by vehicle.
```
wbrannon, line 813: So I think we discussed that we should make it clear this is a software-only message, as in the ticks are extracted from somewhere else and they're sent via these messages.

wbrannon (approved):
Looks good, see my note on the WheelTickInput message.

src/point_one/fusion_engine/messages/measurements.h
adamshapiro0: I think that should be clear after removing the other part? As it says, this message is an input to the device, not an output from the device. The only way for this message to exist is for an external processor to generate it in software.

wbrannon: Ah I see - you're right, looks good to me.

PR #323, opened 2024-06-12
Renamed `UNIXSocketType` enum and added `ALL_PARAMETERS` interface configuration definition.

adamshapiro0:
# New Features
- Added `InterfaceConfigType::ALL_PARAMETERS` and config structs (e.g., `TCPConfig`) to allow bulk configuration of a transport

# Changes
- Renamed `SocketType` to `UNIXSocketType` for clarity


src/point_one/fusion_engine/messages/configuration.h
```diff
@@ -2218,6 +2234,58 @@ inline p1_ostream& operator<<(p1_ostream& stream, MessageRate val) {
   return stream;
 }
 
+/**
+ * @param TCP client/server interface configuration parameters.
+ * @ingroup io_interfaces
+ */
+struct P1_ALIGNAS(4) TCPConfig {
+  TransportDirection direction = TransportDirection::SERVER;
```
axlan, line 2242: Seems like this and the other configs should have an "enabled" field as well.

axlan (changes requested):
This is missing the Python implementation, which is probably going to be non-trivial. Similarly, the Nautilus implementation may be a little complicated since it's going to need to return multiple memory segments, which we've done for wheel speed, but I'm not sure we've done for the interface settings.

adamshapiro0:
@axlan added Python implementation and resolved a few related issues. Please take another look.

src/point_one/fusion_engine/messages/configuration.h
```diff
@@ -2221,6 +2242,62 @@ inline p1_ostream& operator<<(p1_ostream& stream, MessageRate val) {
...
 
+/**
+ * @brief TCP client/server interface configuration parameters.
+ * @ingroup io_interfaces
+ */
+struct P1_ALIGNAS(4) TCPConfig {
+  uint8_t enabled = 1;
+  TransportDirection direction = TransportDirection::SERVER;
+  uint16_t port = 0;
+  char remote_address[64] = {0};
```
anathan, line 2253: its odd to call this remote address if this represents a server and a client.  My personal opinion is that TCPServer and TCPClient should be treated as two different top level entities. There is no such thing as a remote address on a TCP server and if its intended to be the bind_address, it should be named as such.

anathan (approved):
overall the name changes make sense.

axlan (approved):
LGTM

PR #325, opened 2024-06-12
Added stationary position error plot support.

adamshapiro0:
# New Features
- Added `--reference` option to plot pose error compared with a known truth source
  - Currently supports stationary position error only


ryancdougherty (approved):
nice new diagnostic plot

PR #326, opened 2024-06-20
Fix display not showing version.

axlan:
When checking for version messages, don't convert them to numpy so they get parsed correctly.


adamshapiro0 (approved):
LGTM

PR #330, opened 2024-06-28
Add plotting for SV azimuth and elevation angles

wbrannon:
# New Features
- Added new GNSS satellite az/el vs time plot

### DIFF
--- a/python/fusion_engine_client/analysis/analyzer.py
+++ b/python/fusion_engine_client/analysis/analyzer.py
@@ -1148,13 +1148,9 @@ class Analyzer(object):
 
         # Set up the figure.
         figure = make_subplots(
-            rows=4, cols=1,  print_grid=False, shared_xaxes=True,
+            rows=2, cols=1,  print_grid=False, shared_xaxes=True,
             subplot_titles=["Azimuth Angle",
-                            "Elevation Angle", None, None],
-            specs=[[{'rowspan': 2}],
-                   [None],
-                   [{'rowspan': 2}],
-                   [None]])
+                            "Elevation Angle"])
         figure['layout'].update(showlegend=True, modebar_add=['v1hovermode'])
         figure['layout']['xaxis1'].update(title=self.p1_time_label, showticklabels=True)
         figure['layout']['xaxis2'].update(title=self.p1_time_label, showticklabels=True)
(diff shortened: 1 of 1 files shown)


python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -1135,6 +1135,91 @@ def plot_gnss_cn0(self):
...
+            'type': 'buttons',
+            'direction': 'left',
+            'buttons': buttons,
+            'x': 0.0,
+            'xanchor': 'left',
+            'y': 1.1,
+            'yanchor': 'top'
+        }]
+
+        self._add_figure(name='satellite_azimuth_elevation', figure=figure, title='Satellite Azimuth & Elevation')
```
adamshapiro0, line 1221: Title and filename should start with "GNSS" (or "gnss_") so it sorts with the other GNSS plots alphabetically, and add "Vs. Time" to the end similar to other time series plots

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -1135,6 +1135,91 @@ def plot_gnss_cn0(self):
...
+        result = self.reader.read(message_types=GNSSSatelliteMessage, **self.params)
+        data = result[GNSSSatelliteMessage.MESSAGE_TYPE]
+
+        if len(data.p1_time) == 0:
+            self.logger.info('No satellite data available. Skipping azimuth/elevation plot.')
+            return
+
+        # Set up the figure.
+        figure = make_subplots(
+            rows=4, cols=1,  print_grid=False, shared_xaxes=True,
```
adamshapiro0, line 1151: Any particular reason to use 4 rows and a spec below instead of just 2 rows and no `specs=` specifier? Doing it this way complicates things: you need the extra title none entries, the add trace call uses row 3 instead of 2, etc.

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -1135,6 +1135,91 @@ def plot_gnss_cn0(self):
...
+        figure['layout']['yaxis2'].update(title="Degrees")
+
+        # Assign colors to each satellite.
+        data_by_sv = GNSSSatelliteMessage.group_by_sv(data)
+        svs_by_system = defaultdict(set)
+        indices_by_system = defaultdict(list)
+        svs = sorted(list(data_by_sv.keys()))
+        color_by_sv = self._assign_colors(svs)
+
+        for i, sv in enumerate(svs[::-1]):
```
adamshapiro0, line 1171: I don't think you want to iterate backwards here. I'm guessing this was taken from the fix status plot? There we intentionally iterate in reverse order so GPS is at the top of the plot (_highest_ Y axis value), but doing it here would likely put GPS at the bottom of the legend so it doesn't match other plots.

Also doesn't seem like you use `i` at all here so I don't think you need the `enumerate()` call?

adamshapiro0 (approved):
Looks good, just a couple of small things

PR #331, opened 2024-07-23
Add bitmask in `PoseMessage` to signal stationary status of device

wbrannon:
# New Features
- `PoseMessage` now has `flags member, which currently signals whether the device is stationary or not
- Add `Stationary Status` plotting tool

### DIFF
--- a/python/fusion_engine_client/analysis/analyzer.py
+++ b/python/fusion_engine_client/analysis/analyzer.py
@@ -744,8 +744,8 @@ class Analyzer(object):
 
         figure['layout']['xaxis'].update(title=self.p1_time_label)
         figure['layout']['yaxis1'].update(title="Stationary Status",
-                                          ticktext=['Nonstationary', 'Stationary'],
-                                          tickvals=[0, 1])
+                                          ticktext=['Moving', 'Stationary'],
+                                          tickvals=[0, PoseMessage.FLAG_STATIONARY])
 
         time = pose_data.p1_time - float(self.t0)
 
--- a/python/fusion_engine_client/analysis/analyzer.py
+++ b/python/fusion_engine_client/analysis/analyzer.py
@@ -724,6 +724,39 @@ class Analyzer(object):
 
         self._add_figure(name="solution_type", figure=figure, title="Solution Type")
 
+    def plot_stationary_status(self):
+        """!
+        @brief Plot the staionary status over time.
+        """
+        if self.output_dir is None:
+            return
+
+        # Read the pose data.
+        result = self.reader.read(message_types=[PoseMessage], source_ids=self.default_source_id, **self.params)
+        pose_data = result[PoseMessage.MESSAGE_TYPE]
+
+        if len(pose_data.p1_time) == 0:
+            self.logger.info('No pose data available. Skipping solution type plot.')
+            return
+
+        # Set up the figure.
+        figure = make_subplots(rows=1, cols=1, print_grid=False, shared_xaxes=True, subplot_titles=['Stationary Status'])
+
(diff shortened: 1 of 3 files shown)


src/point_one/fusion_engine/messages/solution.h
```diff
@@ -51,7 +54,8 @@ struct P1_ALIGNAS(4) PoseMessage : public MessagePayload {
   /** The type of this position solution. */
   SolutionType solution_type;
 
-  uint8_t reserved = 0;
+  /** Signals device dynamic status (stationary or nonstationary). */
+  uint8_t dynamic_status = NONSTATIONARY;
```
adamshapiro0, line 58: We want this to be a more general purpose flags bitmask, so I'd suggest calling it `flags` and documenting it similar to flags variables in other messages like `WheelSpeedInput` for example. Something like "A bitmask of flags indicating the status of this solution." It doesn't have to be limited to stationary or not, we might also define bits for other things in the future.

For the constants, name it `FLAG_STATIONARY = 0x01` (it's an 8b mask, so don't need all the additional 0s) and get rid of `NONSTATIONARY` since that's not defining a bit, it's just the lack of one particular bit.

The `flags` variable should default to 0x0.

src/point_one/fusion_engine/messages/solution.h
```diff
@@ -42,6 +42,9 @@ struct P1_ALIGNAS(4) PoseMessage : public MessagePayload {
   static constexpr uint8_t MESSAGE_VERSION = 1;
   static constexpr int16_t INVALID_UNDULATION = INT16_MIN;
 
+  /** Set this flag if the device is stationary. */
```
adamshapiro0, line 45: Set ~this flag~ if ("this flag" sounds like something the user should do, not something that the device will do)

python/fusion_engine_client/messages/solution.py
```diff
@@ -152,6 +160,7 @@ def to_numpy(cls, messages):
             'p1_time': np.array([float(m.p1_time) for m in messages]),
             'gps_time': np.array([float(m.gps_time) for m in messages]),
             'solution_type': np.array([int(m.solution_type) for m in messages], dtype=int),
+            'flags': np.array([m.flags for m in messages]),
```
adamshapiro0, line 163: Add `dtype=int` so this isn't cast as floating point

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -724,6 +724,39 @@ def plot_solution_type(self):
...
+        if len(pose_data.p1_time) == 0:
+            self.logger.info('No pose data available. Skipping solution type plot.')
+            return
+
+        # Set up the figure.
+        figure = make_subplots(rows=1, cols=1, print_grid=False, shared_xaxes=True, subplot_titles=['Stationary Status'])
+
+        figure['layout']['xaxis'].update(title=self.p1_time_label)
+        figure['layout']['yaxis1'].update(title="Stationary Status",
+                                          ticktext=['Nonstationary', 'Stationary'],
```
adamshapiro0, line 747: Nonstationary -> Moving

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -724,6 +724,39 @@ def plot_solution_type(self):
...
+                                          tickvals=[0, 1])
```
adamshapiro0, line 748: You should use `PoseMessage.FLAG_STATIONARY` instead of 1 here since below you're just doing a &, not converting it to a bool or something. Technically if this bit value was anything other than 0x1, this wouldn't work correctly.

adamshapiro0 (approved):
Looks good, just small stuff

PR #337, opened 2024-09-04
[FUS-2790] Add Response::UNSUPPORTED_INTERFACE

axlan:
# New Features
- Added separate `UNSUPPORTED_INTERFACE` response code

# Changes
- Issue a compiler warning for missing enum values in `to_string(Response val)`

# Bug Fixes
- Update deprecated Github actions

### DIFF
--- a/python/fusion_engine_client/messages/defs.py
+++ b/python/fusion_engine_client/messages/defs.py
@@ -79,7 +79,7 @@ class Response(IntEnum):
     NO_DATA_STORED = 8
     ## The device is in a state where it can't process the command.
     UNAVAILABLE = 9
-    ## A interface specified in the command is invalid, or unsupported on the target device.
+    ## An interface specified in the command is invalid, or unsupported on the target device.
     UNSUPPORTED_INTERFACE = 10
 
 
--- a/python/fusion_engine_client/messages/defs.py
+++ b/python/fusion_engine_client/messages/defs.py
@@ -79,6 +79,8 @@ class Response(IntEnum):
     NO_DATA_STORED = 8
     ## The device is in a state where it can't process the command.
     UNAVAILABLE = 9
+    ## A interface specified in the command is invalid, or unsupported on the target device.
+    UNSUPPORTED_INTERFACE = 10
 
 
 class MessageType(IntEnum):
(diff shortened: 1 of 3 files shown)


src/point_one/fusion_engine/messages/defs.h
```diff
@@ -414,6 +414,11 @@ enum class Response : uint8_t {
    * The device is in a state where it can't process the command.
    */
   UNAVAILABLE = 9,
+  /**
+   * A TransportType specified in the command is invalid, or unsupported on the
+   * target device.
+   */
+  TRANSPORT_TYPE_ERROR = 10,
```
adamshapiro0, line 421: I'd suggest we call this `UNSUPPORTED_INTERFACE` instead. I think that's a little more clear since this error code should apply for both an unsupported transport (you asked for TCP on a serial-only device, etc.) and also a bad index (you asked for serial port 2000 and there are only 2). Essentially, this applies if any of the fields in the `InterfaceID` structure are wrong, not just the transport type.

Unless you want to have 2 new, separate error codes for invalid index and transport type, where transport type would just be that you gave a valid enum, but one that isn't supported on this type of device? I think invalid enum would just fall under `VALUE_ERROR`.

python/fusion_engine_client/messages/defs.py
```diff
@@ -79,6 +79,8 @@ class Response(IntEnum):
     NO_DATA_STORED = 8
     ## The device is in a state where it can't process the command.
     UNAVAILABLE = 9
+    ## A interface specified in the command is invalid, or unsupported on the target device.
```
adamshapiro0, line 82: A -> An (or The?)

src/point_one/fusion_engine/messages/defs.h
```diff
@@ -414,6 +414,11 @@ enum class Response : uint8_t {
    * The device is in a state where it can't process the command.
    */
   UNAVAILABLE = 9,
+  /**
+   * A interface specified in the command is invalid, or unsupported on the
```
adamshapiro0, line 418: A -> An (or The?)

src/point_one/fusion_engine/messages/defs.h
```diff
@@ -445,9 +450,10 @@ P1_CONSTEXPR_FUNC const char* to_string(Response val) {
       return "No Data Stored";
     case Response::UNAVAILABLE:
       return "Device Unavailable";
-    default:
-      return "Unrecognized";
+    case Response::UNSUPPORTED_INTERFACE:
+      return "Interface ID Error";
```
adamshapiro0, line 454: Not just "Unsupported Interface"?

PR #338, opened 2024-09-25
Add `RESET_GNSS_TIME` and `POSE_RESET` enums for reset requests

wbrannon:


python/fusion_engine_client/messages/control.py
```diff
@@ -258,10 +260,10 @@ class ResetRequest(MessagePayload):
     WARM_START = 0x00000201
 
     ##
-    # Perform a PVT reset: reset all position, velocity, orientation, and time
+    # Perform a Pose reset: reset all position, velocity, and orientation
```
adamshapiro0, line 263: Pose -> pose here and below

src/point_one/fusion_engine/messages/control.h
```diff
@@ -254,10 +256,10 @@ struct P1_ALIGNAS(4) ResetRequest : public MessagePayload {
   static constexpr uint32_t WARM_START = 0x00000201;
 
   /**
-   * Perform a PVT reset: reset all position, velocity, orientation, and time
+   * Perform a Pose reset: reset all position, velocity, and orientation
```
adamshapiro0, line 259: Same

src/point_one/fusion_engine/messages/control.h
```diff
@@ -123,6 +123,8 @@ struct P1_ALIGNAS(4) ResetRequest : public MessagePayload {
   static constexpr uint32_t RESET_POSITION_DATA = 0x00000100;
   /** Delete all saved satellite ephemeris. */
   static constexpr uint32_t RESET_EPHEMERIS = 0x00000200;
+  /** Delete all GNSS time information. */
+  static constexpr uint32_t RESET_TIME = 0x00000004;
```
adamshapiro0, line 127: This is in the wrong place. It should be in the "Runtime State Reset" group above with the other masks in its range.

Should we call this `RESET_GNSS_TIME` for clarity?

adamshapiro0 (approved):
Gotta fix doxygen and other stuff but looks ok otherwise.

PR #340, opened 2024-10-18
Allow programmatic access to `p1_display` internals.

adamshapiro0:
# Changes
- Added a note to `README.md` explaining Windows Python requirements
- Return `Analyzer` from `main()` so it may be accessed further when embedded within another application
- Constrain the 3D displacement Y axis to min 0

### DIFF
--- a/python/fusion_engine_client/analysis/analyzer.py
+++ b/python/fusion_engine_client/analysis/analyzer.py
@@ -814,6 +814,7 @@ class Analyzer(object):
         time_figure.update_layout(title_text=extra_text)
 
         # Plot the data.
+        max_3d_diff_m = [0.0]
         def _plot_data(name, idx, marker_style=None):
             style = {'mode': 'markers', 'marker': {'size': 8}, 'showlegend': True, 'legendgroup': name,
                      'hoverlabel': {'namelength': -1}}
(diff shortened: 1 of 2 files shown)


python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -858,32 +867,32 @@ def plot_stationary_position_error(self, truth_lla_deg):
...
-        self._plot_pose_displacement(title='Position Error', center_ecef_m=truth_ecef_m)
+        return self._plot_pose_displacement(title='Position Error', center_ecef_m=truth_ecef_m)
 
     def plot_pose_displacement(self):
         """!
         @brief Generate a topocentric (top-down) plot of position displacement, as well as plot of displacement over
                time.
         """
-        self._plot_pose_displacement()
+        return self._plot_pose_displacement()
```
wbrannon, line 877: Two things:
1. What's the reason for returning this?
2. Should this be documented?

wbrannon: In response to (1): looking around again, I guess this is consistent with the purpose of this PR.

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -847,6 +850,12 @@ def _plot_data(name, idx, marker_style=None):
         for type, info in _SOLUTION_TYPE_MAP.items():
             _plot_data(info.name, solution_type == type, marker_style=info.style)
 
+        # Set the 3D displacement Y-axis limits to start at 0 and encompass the data.
+        max_y = 1.2 * max_3d_diff_m[0]
+        if max_y == 0.0:
+            max_y = 1.0
+        time_figure['layout']['yaxis1'].update(range=[0, max_y])
```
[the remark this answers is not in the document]
adamshapiro0: Correct

python/fusion_engine_client/analysis/analyzer.py
adamshapiro0: Yep. This isn't really useful internal to this file, but it's handy in our specific use case outside of this file where you want to use the `Analyzer` class instance to do the displacement calculation for you and then do something with that.

PR #342, opened 2024-10-25
Add InputDataWrapperMessage.

axlan:


python/fusion_engine_client/messages/measurements.py
```diff
@@ -1390,3 +1390,53 @@ def to_numpy(cls, messages: Sequence['RawHeadingOutput']):
...
+        return ret
+
+    def __repr__(self):
+        result = super().__repr__()[:-1]
+        result += f', data_type={self.data_type}, data_len={len(self.data)}]'
+        return result
+
+    def __str__(self):
+        return construct_message_to_string(message=self, construct=self.Construct,
+                                           value_to_string={'data': lambda x: f'{len(x)} B payload'},
```
adamshapiro0, line 1438: Suggest the following for easier comparison of data type values:
```
                                           value_to_string={
                                               'data': lambda x: f'{len(x)} B payload',
                                               'data_type': lambda x: f'{x} (0x{x:02x})',
                                           },
```

src/point_one/fusion_engine/messages/measurements.h
```diff
@@ -1261,6 +1261,59 @@ struct P1_ALIGNAS(4) RawHeadingOutput : public MessagePayload {
   float baseline_distance_m = NAN;
 };
 
+/**
+ * @brief A packet of data with an external definition. (@ref
```
adamshapiro0, line 1265: I'd suggest we be careful with the word "packet" since some people might interpret it as framed data. How about:
> A block of incoming sensor data whose definition depends on the value of @ ref data_type.

And then
> This message ... data. The payload is not guaranteed to be aligned to a specific message boundary, or to contain complete messages.

src/point_one/fusion_engine/messages/measurements.h
```diff
@@ -1261,6 +1261,59 @@ struct P1_ALIGNAS(4) RawHeadingOutput : public MessagePayload {
...
+ * ```
+ */
+struct P1_ALIGNAS(4) InputDataWrapperMessage {
+  static constexpr MessageType MESSAGE_TYPE = MessageType::INPUT_DATA_WRAPPER;
+  static constexpr uint8_t MESSAGE_VERSION = 0;
+
+// The MSVC compiler does not allow unaligned bit fields:
+// https://stackoverflow.com/questions/4310728/forcing-unaligned-bitfield-packing-in-msvc
+// unlike Clang and GCC. This means that `uint64_t system_time_cs : 40;` is 5
+// bytes in GCC and 8 bytes in MSVC. This means that on MSVC you'd need to cast
```
adamshapiro0, line 1283: This means that on --> On MSVC, you must cast ...

src/point_one/fusion_engine/messages/measurements.h
```diff
@@ -1261,6 +1261,59 @@ struct P1_ALIGNAS(4) RawHeadingOutput : public MessagePayload {
...
+// @ref system_time_cs_bytes to read and write the timestamp.
+#if defined(_MSC_VER)
+  /**
+   * 5 byte system wall-clock timestamp in centiseconds (hundredths of a
+   * second). Set to POSIX time (time since 1/1/1970) where available.
+   */
+  uint8_t system_time_cs_bytes[5] = {0};
+#else
+  // Default member initializers for bit-fields only available with c++20.
+  InputDataWrapperMessage() : system_time_cs(0) {}
```
adamshapiro0, line 1293: Members should go before functions. It'll mean an extra `#if !defined(_MSC_VER)` but let's move this down below.

src/point_one/fusion_engine/messages/measurements.h
```diff
@@ -1261,6 +1261,59 @@ struct P1_ALIGNAS(4) RawHeadingOutput : public MessagePayload {
...
+#endif
+
+  uint8_t reserved[1] = {0};
+
+  /** Type identifier for the serialized message to follow. */
+  uint16_t data_type = 0;
+
+  /**
+   * The rest of this message contains the wrapped data. The size of the data is
+   * found by subtracting the size of the other fields in this message from the
```
adamshapiro0, line 1309: Maybe write it out as an example: `size_t content_size = header->payload_size_bytes - sizeof(InputDataWrapperMessage)`

adamshapiro0 (approved):
Looks good, just small stuff.

PR #345, opened 2024-12-02
Replaced deprecated `HeadingOutput` messages with new `GNSSAttitudeOutput`.

adamshapiro0:
# New Features
- Added Python `MessagePayload.get_gps_time()` helper function
- Include `gps_time` in numpy output for measurements timestamped in GPS time

# Changes
- Removed deprecated `[Raw]HeadingOutput` messages
- Replaced with `[Raw]GNSSAttitudeOutput`
  - Distinguishes between multi-antenna GNSS attitude _measurements_ and the nav engine's filtered YPR estimate
  - Adds YPR and baseline distance standard deviation
  - Removes redundant heading angle fields


adamshapiro0:
@anathan FYI

PR #346, opened 2024-12-04
Removed deprecated HeadingBias config in favor of GNSS aux lever arm specification.

adamshapiro0:
# New Features
- Added new `GNSS_AUX_LEVER_ARM` (secondary antenna) configuration setting

# Changes
- Removed deprecated `HEADING_BIAS` configuration setting

# Fixes
- Corrected Python `RawGNSSAttitudeOutput` yaw/pitch printout


axlan (approved):
In general seems fine. I would have some concern for backwards compatibility, but I don't think that should be a concern for a feature that probably hasn't had much active use (I don't think we need to worry about supporting devices running older firmware with this updated FE client library).

PR #347, opened 2024-12-09
Added new `p1_capture` utility, and simplified TCP/UDP/serial example apps.

adamshapiro0:
# New Features
- Added new `p1_capture` tool to display incoming FE messages in real time and capture to disk
  - Combines logic formerly in example `tcp_client.py`, `udp_client.py`, and `serial_client.py` applications
  - Adds new `--summary` mode to display a summary of device and FusionEngine message status in real time similar to `p1_print --summary`

# Changes
- Added device type, ID, and software version to `p1_print --summary`
- Drastically simplified `tcp_client.py`, `udp_client.py`, and `serial_client.py` example apps
  - Use `p1_capture` for fully-featured real time capture and display instead

# Fixes
- Use serial/socket read timeouts so we print _some_ console update if there's no incoming data

### DIFF
--- a/python/fusion_engine_client/applications/p1_capture.py
+++ b/python/fusion_engine_client/applications/p1_capture.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python3
 
+from typing import Union
+
 from collections import defaultdict
 from datetime import datetime
 import math
@@ -35,7 +37,7 @@ from ..utils.argument_parser import ArgumentParser, ExtendedBooleanAction
 _logger = logging.getLogger('point_one.fusion_engine.applications.p1_capture')
 
 
-def create_transport(descriptor: str):
+def create_transport(descriptor: str) -> Union[socket.socket, serial.Serial]:
     m = re.match(r'^tcp://([a-zA-Z0-9-_.]+)?:([0-9]+)$', descriptor)
     if m:
         transport = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
--- a/python/fusion_engine_client/applications/p1_capture.py
+++ b/python/fusion_engine_client/applications/p1_capture.py
@@ -2,7 +2,6 @@
 
 from typing import Union
 
-from collections import defaultdict
 from datetime import datetime
 import math
 import os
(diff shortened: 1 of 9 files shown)


python/fusion_engine_client/applications/p1_print.py
```diff
@@ -36,7 +38,10 @@ def add_print_format_argument(parser, *arg_names):
              "- oneline-binary-payload - Like `oneline-binary`, but exclude the message header from the binary")
 
 
-def print_message(header, contents, offset_bytes=None, format='pretty', bytes=None):
+def print_message(header, contents, offset_bytes=None, format='pretty', bytes=None, logger=None):
```
axlan, line 41: Code organization wise, it seems like it would be a little clearer to break out these display functions into their own file. I also think it would be less confusing not to pass in a logger, but just use the normal logging level control mechanisms for this shared code. I assumed you just needed to do that since this logger had "application" in it's name?

python/fusion_engine_client/applications/p1_capture.py
```diff
@@ -0,0 +1,265 @@
...
+from .p1_print import \
+    DeviceSummary, add_print_format_argument, print_message, print_summary_table
+from ..parsers import FusionEngineDecoder
+from ..utils import trace as logging
+from ..utils.argument_parser import ArgumentParser, ExtendedBooleanAction
+
+_logger = logging.getLogger('point_one.fusion_engine.applications.p1_capture')
+
+
+def create_transport(descriptor: str):
```
axlan, line 38: I think this file would benefit from type checks, since it's not very obvious what set of functions are supported on the returned transport. So returning a Union here combined with the isinstance checks should catch any mistakes.

axlan (approved):
LGTM

python/fusion_engine_client/applications/p1_print.py
adamshapiro0: I can move these to a separate file. I passed the logger in since from the user's perspective these prints come from the application and it was odd to have it print from a different logging path. If I'm moving this to a separate file it's probably ok to remove this.

PR #349, opened 2024-12-14
Capture raw, unfiltered data by default in `p1_capture`.

adamshapiro0:
# Changes
- Capture in `raw` format by default in `p1_capture`

# Fixes
- Fixed `p1_capture --output-format` option


axlan (approved):
LGTM

PR #354, opened 2025-01-31
Add config type for enabling profiling.

axlan:


src/point_one/fusion_engine/messages/configuration.h
```diff
@@ -255,6 +255,13 @@ enum class ConfigType : uint16_t {
    */
   USER_DEVICE_ID = 301,
 
+  /**
+   * A bitmask indicating which profiling features are enabled.
+   *
+   * Payload format: `uint8_t` (see @ref sat_type_masks)
```
adamshapiro0, line 261: This sat_type_masks reference seems wrong?

adamshapiro0, line 261: Should this be a uint32 for expandability?

src/point_one/fusion_engine/messages/configuration.h
```diff
@@ -255,6 +255,13 @@ enum class ConfigType : uint16_t {
...
+   */
+  PROFILING_ENABLED = 310,
```
adamshapiro0, line 263: Maybe `PROFILING_MASK` instead? Enabled sort of implies a single bool instead of multiple things

python/fusion_engine_client/messages/configuration.py
```diff
@@ -886,6 +887,14 @@ class UserDeviceID(_conf_gen.StringVal):
     pass
 
 
+@_conf_gen.create_config_class(ConfigType.PROFILING_ENABLED, _conf_gen.UInt8Construct)
```
adamshapiro0, line 890: Reminder to change to 32 if you change below

adamshapiro0 (approved):
Looks good

src/point_one/fusion_engine/messages/configuration.h
axlan: I considered this, but:
1. The device side implementation is much simpler when there's a 1-1 mapping
2. I don't want the actual profiling details to be part of the public standard which also is simpler if its a 1-1 mapping
3. It's trivial to add a new type if we want to have a more transparent setting

PR #357, opened 2025-03-04
Cleaned up p1_display log time table.

adamshapiro0:
# Fixes
- Corrected formatting and timestamps displayed on `p1_display` index page
- Fixed message count in `MessageData` object `str()` output after numpy conversion


ryancdougherty (approved):
All looks good.

PR #360, opened 2025-03-18
Fixed `p1_display` reset timing plot crash.

adamshapiro0:
# Changes
- Renamed `MessageData.messages_bytes` list to `message_bytes` to match `message_index` list

# Fixes
- Don't delete the `message_index` list during numpy conversion if `return_message_index=True`


axlan (approved):
LGTM

PR #363, opened 2025-03-24
Fixed scroll zoom support on plotly map.

adamshapiro0:
# Fixes
- Fixed plotly map scroll wheel zoom setting


ryancdougherty (approved):
So simple.

PR #368, opened 2025-04-17
Handle non-OK `ConfigResponseMessages` gracefully in Python.

adamshapiro0:
# Fixes
- Check the response type and data size before decoding the config response payload


axlan (approved):
LGTM

PR #369, opened 2025-04-17
Added common helper function for defining log search CLI arguments.

adamshapiro0:
# Changes
- Added common helper function for defining log search CLI arguments

# Fixes
- Fixed `p1_display` default output directory when reading from a `.p1log` file that's not within a P1 log directory


axlan (approved):
LGTM

PR #370, opened 2025-05-14
Fixed displacement plot median position calculation.

adamshapiro0:
# Fixes
- Fixed displacement plot median position calculation
- Include baseline distance/std dev in raw GNSS attitude printout


axlan (approved):
LGTM

PR #396, opened 2026-04-09
Fixed Python `TransportType` name conflict, and added message byte count printout.

adamshapiro0:
# Changes
- Print the total byte count for each message type in summary table

# Fixes
- Fixed name conflict between `TransportType` in `configuration.py` and `transport_utils.py`
- Fixed missing `;` in injected JS plotting code


wbrannon (approved):
Looks good.

PR #399, opened 2026-04-21
Fixed path specifier error for output files.

adamshapiro0:
# Fixes
- Fixed path specifier error for output files.


bhers4 (approved):
Looks good to me!

PR #402, opened 2026-04-22
Combined `p1_capture`, `p1_filter`, and `p1_print` utilities.

adamshapiro0:
# Changes
- Refactored `p1_capture` to combine capabilities with `p1_filter` and `p1_print` into a single tool
- Added message type filter invert option to `MixedLogReader` and `FileIndex`

# Fixes
- Fixed `FileIndex` time range filtering if the end time is earlier than the start of the data and the first messages do not have P1 timestamps
- Expand `~/` when parsing input/output file paths
- Fixed Python 3.8/3.9 compatibility issues


python/fusion_engine_client/applications/p1_capture.py
```diff
@@ -5,72 +5,756 @@
...
+            self.include_input_data_wrapper = True
+            if self.wrapped_data_format == 'auto':
+                self.wrapped_data_format = 'content'
+
+    def _init_input_data_type_filter(self):
+        # For InputDataWrapper messages, if the user specified desired data types, limit the output to only those.
+        if self.options.wrapped_data_type is not None:
+            try:
+                self.input_data_types = InputDataType.find_matching_values(self.options.wrapped_data_type, prefix='M_TYPE_',
+                                                                      print_func=_logger.error)
```
bhers4: Minor nit: Make `print_func` right under the `self.options`

python/fusion_engine_client/applications/p1_capture.py
```diff
@@ -5,72 +5,756 @@
...
+    def _init_source_id_filter(self):
+        # If the user specified a set of source IDs, limit messages to only those sources.
+        if self.options.source_identifier is not None:
+            try:
+                self.source_ids = set([int(s) for s in self.options.source_identifier])
+            except ValueError:
+                _logger.error('Source identifiers must be integers.')
+                sys.exit(1)
+
+    def _init_time_range_filter(self):
```
bhers4: Minor nit: Add (self) -> None: so at a glance users know its not returning anything. Apply throughout.

python/fusion_engine_client/applications/p1_capture.py
```diff
@@ -5,72 +5,756 @@
...
+        else:
+            if self.options.log_timestamp_source and self.options.log_timestamp_source != 'user-sw':
+                _logger.error(
+                    f'--log-timestamp-source={self.options.log_timestamp_source} is not supported. Only "user-sw" '
+                    f'timestamps are supported on non-socket captures.')
+                sys.exit(1)
+
+            set_read_timeout(self.input_transport, self.read_timeout_sec)
+
+    def process_input(self):
```
bhers4: Add a method comment.

python/fusion_engine_client/applications/p1_capture.py
```diff
@@ -5,72 +5,756 @@
...
+
+                if self.options.max is not None and self.messages_sent == self.options.max:
+                    return False
+
+            if self.time_range is not None and self.time_range.in_range_ended():
+                return False
+
+        return True
+
+    def _apply_filters(self, header: MessageHeader, message: Union[MessagePayload, bytes]):
```
bhers4: Add type hinting, this returns a bool.

python/fusion_engine_client/parsers/file_index.py
```diff
@@ -293,6 +293,46 @@ def save(self, index_path: str, data_path: str):
                 os.remove(index_path)
             raw_data.tofile(index_path)
 
+    def get_message_types(self, message_types: Union[int, IntEnum, MessagePayload, set, list, tuple],
+                          invert: bool = False) -> 'FileIndex':
```
bhers4, line 297: Why not -> FileIndex? Why the 'FileIndex' in quotations?

python/fusion_engine_client/applications/p1_capture.py
```diff
@@ -5,72 +5,756 @@
...
+            #   -m Type1,Type2 -m Type3
+            #   -m Type*
+            try:
+                self.message_types = MessagePayload.find_matching_message_types(self.options.message_type)
+                if len(self.message_types) == 0:
+                    # find_matching_message_types() will print an error.
+                    sys.exit(1)
+            except ValueError as e:
+                _logger.error(str(e))
+                _logger.debug(traceback.print_exception(e))
```
bhers4: From the intellisense, `print_exception` returns none, so the `_logger.debug` does nothing. So you could change it to just `traceback.print_exception(e)` without any logger. Check for this throughout.

python/fusion_engine_client/applications/p1_capture.py
```diff
@@ -5,72 +5,756 @@
...
+                        self._print_display(now)
+
+                if finished:
+                    break
+        except (BrokenPipeError, KeyboardInterrupt) as e:
+            # User hit Ctrl-C -- done processing.
+            pass
+
+        # Close the transport.
+        self.input_transport.close()
```
bhers4, line 427: `_configure_input` already calls: `self.input_transport.input.close()`. Is that an issue with a potential double close?

python/fusion_engine_client/applications/p1_capture.py
```diff
@@ -5,72 +5,756 @@
...
+
+        # Message summary table.
+        self._print("")
+        print_summary_table(self.device_summary)
+
+    def _print(self, msg, *args, **kwargs):
+        if self.quiet:
+            pass
+        else:
+            _logger.info(msg, *args, **kwargs)
```
bhers4, line 605: Why not:
```
if not self.quiet:
   .....
```

python/fusion_engine_client/applications/p1_capture.py
```diff
@@ -5,72 +5,756 @@
...
+        # a time to reduce latency. Otherwise, if we're just displaying stuff or writing to disk, we'll read more data
+        # at a time to be more efficient.
+        is_real_time = (self.output_transport is not None and
+                        (not isinstance(self.output_transport, FileTransport) or self.output_transport.is_stdout))
+        if is_real_time:
+            self.read_timeout_sec = 1.0
+            self.read_size_bytes = 64
+        else:
+            self.read_timeout_sec = 1.0
+            self.read_size_bytes = 1024
```
bhers4, line 302: move `read_timeout_sec` out of the if/else.

bhers4 (approved):
Looks good to me!

python/fusion_engine_client/applications/p1_capture.py
adamshapiro0: It should be safe since we only do that for file inputs, and they support double-close. I checked all the other transports and the only one that would not work is `socket.socket`. I'll make a separate PR to add a `SocketTransport` wrapper to protect that too.

python/fusion_engine_client/applications/p1_capture.py
adamshapiro0: I actually did this intentionally so it's clear that we really want to use this timeout value for each case. If we move it outside, someone might think it just got forgotten.

python/fusion_engine_client/parsers/file_index.py
adamshapiro0: In older versions of Python, when a member function returns a ref to the current class, you can't reference the class in the type hint since the class isn't fully defined when the interpreter is reading the type hint.

In Python 3.10 they added `from __future__ import annotations` that allows doing just `-> FileIndex`, but we currently still support older than that. In 3.11 they also added a newer way `from typing import Self` --> `-> Self`. Both methods require an extra import.

python/fusion_engine_client/applications/p1_capture.py
adamshapiro0: Ah whoops. I want to print with the logger since we may redirect the output depending on the `--output` setting. I meant to call `_logger.debug(traceback.format_exception(e))`, but actually there's a better option if we want to maintain backwards compatibility: `_logger.debug('', exc_info=e)`. The API for `format_exception()` changed in 3.10, before that you have to pass in multiple arguments. `exc_info=e` has worked since 3.5.

PR #403, opened 2026-04-23
 Added `SocketTransport` wrapper class.

adamshapiro0:
# Changes
- Added `SocketTransport` wrapper class to protect against multiple `close()` calls


bhers4 (approved):
Looks good to me! Should make sure we don't double close a socket in the `p1_capture` tool by accident.

PR #407, opened 2026-05-19
Fixed --log-type argument definition.

adamshapiro0:
# Fixes
- Fixed type error in `--log-type` argument definition


bhers4 (approved):
Looks good to me! Good catch!

PR #408, opened 2026-05-20
Fixed missing sky plot entries for short-lived signals.

adamshapiro0:
# Fixes
- Fixed missing sky plot entries for short-lived signals


wbrannon (approved):
Looks good.

PR #409, opened 2026-05-22
Added PoseMessage flag for GNSS receiver solution output.

adamshapiro0:
# New Features
- Added new `PoseMessage::FLAG_RECEIVER_SOLUTION`
  - Separate nav engine/GNSS receiver solutions in solution type plot and map

# Changes
- Plot 3D speed in pose vs time plot
- Set `namelength=-1` on all figures


bhers4 (approved):
Looks good to me!

PR #412, opened 2026-07-01
Corrected altitude value used in pose displacement plot.

adamshapiro0:
# Fixes
- Fixed incorrect use of latitude instead of altitude in pose displacement plot


ryancdougherty (approved):
:+1:

PR #416, opened 2026-07-21
Added in plotting of secondary antenna data if it exists

bhers4:
# Changes
- Added in plotting of the secondary data for the C/N0 plot, Signal Status plot, and Sky plot if a secondary antenna exists.

### DIFF
--- a/python/fusion_engine_client/analysis/analyzer.py
+++ b/python/fusion_engine_client/analysis/analyzer.py
@@ -1804,6 +1804,14 @@ figure.on('plotly_hover', function(data) {{
         params = copy.deepcopy(self.params)
         params['return_numpy'] = False
 
+        # Some devices output GNSSSignalsMessage data for both the primary and secondary GNSS antennas. Exclude the
+        # secondary antenna's data so it is not plotted on top of the primary antenna's. Source ID 1 is a legacy
+        # secondary antenna identifier predating the SourceIdentifier reserved ranges, used by some older devices
+        # instead of SECONDARY_GNSS_ANTENNA (301).
+        secondary_source_ids = {1, SourceIdentifier.SECONDARY_GNSS_ANTENNA}
+        available_source_ids = self.reader.get_available_source_ids()
+        params['source_ids'] = available_source_ids - secondary_source_ids
+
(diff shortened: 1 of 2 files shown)


python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -1782,31 +1795,57 @@ def _count_selected(selected_p1_times, return_nonzero_time=False):
...
+        else:
+            result = self.reader.read(message_types=[GNSSSignalsMessage], **params)
+            data = result[GNSSSignalsMessage.MESSAGE_TYPE]
 
         # We store the result now, even if there were no GNSSSignalsMessage messages. That way if we also don't have any
         # GNSSSatelliteMessage messages below, we'll have _something_ to return and we won't try to reload from disk on
         # each call to this function.
-        self._gnss_signals_data = data
-        self._gnss_signals_data.using_legacy_satellite_message = False
+        self._gnss_signals_data[secondary] = data
```
adamshapiro0: We can do this as a follow-up if we want in interest of time, but it would be nice to generalize this as keying by `source_id` instead of a `secondary` bool, and auto-detect which sources are present. Then the plot functions, or `main()` calling them, could just loop over the available sources.

python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -1782,31 +1795,57 @@ def _count_selected(selected_p1_times, return_nonzero_time=False):
...
+        data.using_legacy_satellite_message = False
 
         # If we don't have any GNSSSignalsMessages, see if we have the legacy GNSSSatelliteMessage and fall back to
-        # that.
-        if len(data.messages) == 0:
+        # that. The legacy message predates dual-antenna support, so there's no secondary antenna data to fall back
+        # to.
```
adamshapiro0: Current software can still produce the legacy `GNSSSatelliteMessage` message, and will generate them for both antennas, so we should treat this the same way as `GNSSSignalsMessage`

adamshapiro0 (approved):
Looks good. We should think about generalizing this more as a follow-on step.

PR #417, opened 2026-07-21
Added time-varying reference data support.

adamshapiro0:
# New Features
- Added `--reference` option to plot pose error using another device's `*.p1log` file as truth


wbrannon (approved):
Looks good.

PR #421, opened 2026-07-29
Plot reference position and position error/std dev on map.

adamshapiro0:
# New Features
- Plot the reference data position as separate traces on the map
- Include position error and position std dev in map hover text


bhers4 (approved):
Looks good to me!

PR #423, opened 2026-07-30
Upgrade required plotly version to >=6.9.0

ryancdougherty:
# Changes
- Upgrade required plotly version to >=6.9.0

# Fixes
- Fixed map rendering bug under new plotly version caused by change in trace serialization
- Fixed graph div object name


adamshapiro0 (approved):
LGTM

PR #431, opened 2026-08-19
Fixed time handling issues and missing map street names.

adamshapiro0:
# Fixes
- Fixed `FileIndex` loading if time hint is specified but start/stop not specified
- Fixed `DataLoader` handling of messages containing both P1 and system timestamps
- Fixed dtype on `MeasurementDetails` GPS and system time numpy arrays
- Fixed missing street names on `p1_display` map

### DIFF
--- a/python/fusion_engine_client/analysis/data_loader.py
+++ b/python/fusion_engine_client/analysis/data_loader.py
@@ -441,8 +441,12 @@ class DataLoader(object):
         # fast reading, messages with system times may have their index entry timestamps set to NAN since A) they can
         # occur in a log before P1 time is established, and B) there's not necessarily a direct way to convert between
         # system and P1 time.
-        p1_time_messages_requested = any([t in messages_with_p1_time for t in needed_message_types])
-        system_time_messages_requested = any([t in messages_with_system_time for t in needed_message_types])
+        requested_messages_with_p1_time = any([t in messages_with_p1_time for t in needed_message_types])
+        requested_messages_with_system_time = any([t in messages_with_system_time for t in needed_message_types])
+        all_system_time_messages_have_p1_time = all([t in messages_with_p1_time
+                                                     for t in needed_message_types if t in messages_with_system_time])
+        requested_messages_with_only_system_time = (requested_messages_with_system_time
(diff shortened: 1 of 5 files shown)


python/tests/test_file_index.py
```diff
@@ -354,9 +354,14 @@ def _lower_bound(time):
     # If the log does not contain P1 time, slicing it by time is not supported.
     with pytest.raises(IndexError):
         sliced_index = index[1.0:]
```
bhers4: Should we delete this? Seems like this just is overwritten by the last line?

python/fusion_engine_client/analysis/data_loader.py
```diff
@@ -441,8 +441,12 @@ def _read(self,
         # fast reading, messages with system times may have their index entry timestamps set to NAN since A) they can
         # occur in a log before P1 time is established, and B) there's not necessarily a direct way to convert between
         # system and P1 time.
-        p1_time_messages_requested = any([t in messages_with_p1_time for t in needed_message_types])
-        system_time_messages_requested = any([t in messages_with_system_time for t in needed_message_types])
+        requested_messages_with_p1_time = any([t in messages_with_p1_time for t in needed_message_types])
+        requested_messages_with_system_time = any([t in messages_with_system_time for t in needed_message_types])
+        all_system_time_messages_have_p1_time = all([t in messages_with_system_time
```
bhers4: shouldn't this be:
```
all_system_time_messages_have_p1_time = all([t in messages_with_p1_time
                                             for t in needed_message_types if t in messages_with_system_time])
```

bhers4 (approved):
Looks good to me, just 1 thing to check!

python/tests/test_file_index.py
adamshapiro0: These are really just to call time range indexing in different ways and verify that they all raise exceptions. We don't actually need the result.

I'll remove the `sliced_index =` and also use separate `with` statements for each one of these calls just to be clear

python/fusion_engine_client/analysis/data_loader.py
adamshapiro0: Yes, will fix

PR #434, opened 2026-09-16
Fixed skyplot crash and source ID 0 warnings.

adamshapiro0:
# Changes
- If body velocity is not available and PoseAux is, plot ENU velocity
- Set relative t0 to the start of reference data when available

# Fixes
- Fixed skyplot crash if message timestamps are out of order
- Don't warn about source ID 0 for logs using recently defined GNSS source IDs


python/fusion_engine_client/analysis/analyzer.py
```diff
@@ -608,9 +657,23 @@ def plot_pose(self):
...
+        # If there's no body velocity available, see if we have PoseAux messages and plot ENU velocity instead.
+        vel_mps = pose_data.velocity_body_mps
+        vel_std_mps = pose_data.velocity_std_body_mps
+        is_body_vel = True
+
+        if np.all(np.isnan(vel_mps)):
+            result = self.reader.read(message_types=[PoseAuxMessage], source_ids=self.default_source_id, **self.params)
+            pose_aux_data = result[PoseAuxMessage.MESSAGE_TYPE]
+            if len(pose_aux_data.p1_time) != 0:
+                vel_mps = pose_aux_data.velocity_enu_mps
```
bhers4, line 669: I think its probably fine but is there any risk of the PoseAux messages and Pose messages not having the same amount of entries causing issues? Just notice time comes from Pose messages but vel if its all NAN comes from PoseAux messages.

bhers4 (approved):
Looks good to me!

PR #436, opened 2026-09-16
Fix Python `ExportDataMessage` serialization and handling of truncated command/response events.

adamshapiro0:
# Fixes
- Handle `StreamError` gracefully when decoding `EventNotificationMessages`
- Detect truncated commands captured within `EventNotificationMessages`
- Fixed padding in Python `ExportDataMessage` class


python/fusion_engine_client/messages/device.py
```diff
@@ -256,23 +260,41 @@ def _flags_to_string(value):
...
+                    len(self.event_description) < (MessageHeader.calcsize() + header.payload_size_bytes)):
+                    message_repr += ' <Truncated>'
+                else:
+                    message_cls = MessagePayload.get_message_class(header.message_type)
+                    if message_cls is not None:
+                        try:
+                            message = message_cls()
+                            message.unpack(buffer=self.event_description, offset=header.calcsize())
+                            message_repr = repr(message)
+                        except Exception:
```
bhers4, line 293: Should we add any debug prints here? Or would the `<Truncated>` be enough?

src/point_one/fusion_engine/messages/device.h
```diff
@@ -267,6 +267,13 @@ struct P1_ALIGNAS(4) EventNotificationMessage : public MessagePayload {
   static constexpr MessageType MESSAGE_TYPE = MessageType::EVENT_NOTIFICATION;
   static constexpr uint8_t MESSAGE_VERSION = 0;
 
+  /**
+   * For command/response messages, this flag indicates that the command could
+   * not be serialized completely within this message, and the payload was
+   * truncated.
+   */
+  constexpr uint64_t FLAG_COMMAND_TRUNCATED = 0x80000000'00000000ULL;
```
bhers4, line 275: Does this need to be static? And do we need a matching `'`? Just verify this compiles properly.

bhers4 (approved):
Looks good to me! Just one small thing to check?

bhers4:
Oh yeah look at the bazel builds are failing with:
```
src/point_one/fusion_engine/messages/device.h:275:3: error: non-static data member 'FLAG_COMMAND_TRUNCATED' declared 'constexpr'
  275 |   constexpr uint64_t FLAG_COMMAND_TRUNCATED = 0x80000000'00000000ULL;
```

